PR #13, opened 2015-02-02
Fixed JSON parser issue: missing NUL byte in parse_number() in some code paths.

cfallin:


haberman:
This PR seems to be for much more than just this small fix?

PR #19, opened 2015-05-15
Bugfix: seeded_alloc() should not realloc() user-provided memory.

cfallin:
When the user provides memory to `upb_seededalloc_init()`, the memory continues to belong to the user, and the user is responsible for freeing the memory after the allocator is eventually destroyed. (This is compatible with, for example, user-provided stack-allocated memory.) However, the current implementation passes the user-provided block to the fallback allocator, and the default fallback allocator attempts to `realloc()` the memory. This results in a libc error and/or heap corruption if the allocations ever exceed the initial user-provided block.

### DIFF
--- a/upb/env.c
+++ b/upb/env.c
@@ -211,6 +211,8 @@ static size_t align_up(size_t size) {
 
 UPB_FORCEINLINE static void *seeded_alloc(void *ud, void *ptr, size_t oldsize,
                                           size_t size) {
+  UPB_UNUSED(ptr);
+
   upb_seededalloc *a = ud;
   size = align_up(size);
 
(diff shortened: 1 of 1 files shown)


cfallin:
Hmm. Neither Travis failure reproduces locally, and neither looks related. Something strange is going on. We may just want to stay on the old version of upb until `seeded_alloc` issues are worked out...

haberman:
Don't worry about the Travis failures, they are present in the current build also (I'm working on fixing it as we speak).

upb/env.c
```diff
@@ -224,7 +226,9 @@ UPB_FORCEINLINE static void *seeded_alloc(void *ud, void *ptr, size_t oldsize,
   } else {
     // Slow path: fallback to other allocator.
     a->need_cleanup = true;
-    return a->alloc(a->alloc_ud, ptr, oldsize, size);
+    // N.B.: do not pass initial block, since it was user-provided; otherwise,
```
haberman: Ah, actually I think maybe the correct predicate here is:

```
char *chptr = ptr;
void *from = (chptr >= a->mem_base && chptr < a->mem_limit) ? NULL : ptr;
```

Because we ought to pass ptr in the case that it is *not* from the initial block.

haberman:
Sorry, I didn't expect this to require you to chase down the rabbit-hole of upb!

upb/env.c
cfallin: Good catch, updated, thanks!

PR #21, opened 2015-05-17
Restructure tables for C89 port and smaller size.

haberman:
Changes the data layout of tables slightly so that string
keys are prefixed with their size, rather than the size
being inline in the table itself.

This has a few benefits:

1. inttables shrink a bit, because there is no longer a wasted
   and unused size field sitting in them.

2. This avoids the need to have a union in the table.  This is
   important for an impending C89 port of upb, since C89 has
   literally no way of statically initializing a non-first union
   member.


haberman:
@cfallin want to take a look?

tests/test_table.cc
```diff
@@ -62,6 +62,8 @@ void test_strtable(const vector<std::string>& keys, uint32_t num_to_insert) {
       upb_strtable_next(&iter)) {
     const char *key = upb_strtable_iter_key(&iter);
     std::string tmp(key, strlen(key));
+    std::string tmp2(key, upb_strtable_iter_keylength(&iter));
+    ASSERT(tmp == tmp2);
```
cfallin: Maybe `ASSERT(strlen(key) == upb_strtable_iter_keylength(&iter))` here to avoid extra copy and data comparison?

upb/table.int.h
```diff
@@ -152,33 +152,46 @@ FUNCS(fptr,     fptr,         upb_func*,    UPB_CTYPE_FPTR);
...
-#endif
-// TODO(haberman): C++
-#define UPB_TABKEY_NONE {0}
+#define UPB_TABKEY_NUM(n) n
+#define UPB_TABKEY_NONE 0
+// The preprocessor isn't quite powerful enough to turn the compile-time string
+// length into a byte-wise string representation, so code generation needs to
+// help it along.
+//
+// "len1" is the low byte and len4 is the high byte.  For big endian we'll need
```
cfallin: Is there a macro we can test and `#error` on for now to make sure we don't compile on big-endian until this is implemented?

upb/table.int.h
```diff
@@ -152,33 +152,46 @@ FUNCS(fptr,     fptr,         upb_func*,    UPB_CTYPE_FPTR);
...
+#define UPB_TABKEY_STR(len1, len2, len3, len4, strval) \
+    (uintptr_t)(len1 len2 len3 len4 strval)
+
+// Either:
+//   1. an actual integer key, or
+//   2. a pointer to a string prefixed by its uint32_t length, owned by us.
+//
+// ...depending on whether this is a string table or an int table.  We would
+// make this a union of those two types, but C89 doesn't support statically
+// initializing a non-first union member.
```
cfallin: An alternative idea (maybe you've already tried this): can this continue to be a union, but we simply use initializers that cast values to the type of the first union member, which would be a `uintptr_t`?

cfallin:
Looks good overall; see comments above for possible alternative suggestion and a few nits.

upb/table.int.h
haberman: I thought about that, but it seemed slightly more sketchy from an undefined behavior perspective (there is even less support for type punning in C89 than C99, IIRC), and it doesn't buy that much convenience, since there's only one place that needs to cast it back to the pointer type.

cfallin: Seems reasonable, thanks!

upb/table.int.h
haberman: I implemented some basic endian support and then defined the big-endian version of this macro.

tests/test_table.cc
haberman: Haha, sure, though I don't usually worry too much about this level of optimization in unit tests. :)

haberman:
Thanks, PTAL.

upb/refcounted.c
```diff
@@ -726,6 +726,18 @@ static void freeobj(upb_refcounted *o) {
...
                          const void *owner) {
+#ifndef NDEBUG
+  // Endianness check.  This is unrelated to upb_refcounted, it's just a
+  // convenient place to put the check that we can be assured will run for
+  // basically every program using upb.
+  const int x = 1;
+#ifdef UPB_BIG_ENDIAN
+  assert(*(char*)&x != 1);
+#else
+  assert(*(char*)&x == 1);
```
cfallin: This will run on every descriptor object creation, right? Maybe not too big a deal, but I wonder if it could be made a compile-time check.

Actually, scratch that. gcc is apparently smart enough to do constant-prop through a typecast -- this compiles down to a `ret` instruction with `-O3`:
    void f() {
      const int x = 1;
      assert(*(char*)&x == 1);
    }

Maybe add a comment to allay equivalent worries from future readers that this should be optimized away?

upb/refcounted.c
```diff
@@ -726,6 +726,18 @@ static void freeobj(upb_refcounted *o) {
 bool upb_refcounted_init(upb_refcounted *r,
                          const struct upb_refcounted_vtbl *vtbl,
                          const void *owner) {
+#ifndef NDEBUG
+  // Endianness check.  This is unrelated to upb_refcounted, it's just a
+  // convenient place to put the check that we can be assured will run for
+  // basically every program using upb.
+  const int x = 1;
+#ifdef UPB_BIG_ENDIAN
+  assert(*(char*)&x != 1);
```
cfallin: `!= 1` --> `== 0` ?

upb/refcounted.c
haberman: Sure I can add a note. Though we do waaaay more expensive runtime checks than this when NDEBUG isn't enabled. I am kind of impressed that you are out-performance-paranoia'ing me. :)

Actually if this is truly optimized away, I'm tempted to leave it in for release builds too. The only problem is that I'm not sure what it ought to do if the check fails. It is such a shame that this can't be a compile-time assert!

upb/refcounted.c
haberman: Done.

cfallin:
I see your "Done" comments but `2cff15d` is still the latest commit; did the update get lost somehow?

haberman:
Sorry I just now pushed the update.

PR #34, opened 2015-07-30
Changed parser semantics around skipping.

haberman:
Prior to this change:

  parse(buf, len) -> len + N

...would indicate that the next N bytes of the input are not
needed, *and* would advance the decoding position by this
much.

After this change:

  parse(buf, len) -> len + N
  parse(NULL, N) -> N

...can be used to achieve the same thing.  But skipping the
N bytes is not explicitly performed by the user.  A user that
doesn't want/need to skip can just say:

  parsed = parse(buf, len);

  if (parsed < len) {
    // Handle suspend, advance stream by "parsed".
  } else {
    // Stream was advanced by "len" (even if parsed > len).
  }

Updated unit tests to test this new behavior, and refactored
test utility code a bit to support it.

### DIFF
--- a/tests/test_util.h
+++ b/tests/test_util.h
@@ -8,59 +8,90 @@
 #include <stdio.h>
 #include <math.h>
 #include "tests/upb_test.h"
+#include "upb/env.h"
 #include "upb/sink.h"
 
 upb::BufferHandle global_handle;
 
-// Puts a region of the given buffer [start, end) into the given sink (which
-// probably represents a parser.  Can gracefully handle the case where the
-// parser returns a "parsed" length that is less or greater than the input
-// buffer length, and tracks the overall parse offset in *ofs.
-//
-// Pass verbose=true to print detailed diagnostics to stderr.
-bool parse_buffer(upb::BytesSink* sink, void* subc, const char* buf,
-                  size_t start, size_t end, size_t* ofs,
-                  upb::Status* status, bool verbose) {
-  start = UPB_MAX(start, *ofs);
+class VerboseParserEnvironment {
+ public:
+  VerboseParserEnvironment(bool verbose) : verbose_(verbose) {
+    env_.ReportErrorsTo(&status_);
+  }
 
-  if (start <= end) {
-    size_t len = end - start;
+  void Reset(const char *buf, size_t len, bool may_skip) {
+    buf_ = buf;
+    len_ = len;
+    ofs_ = 0;
+    skip_until_ = may_skip ? 0 : -1;
+    status_.Clear();
+  }
+
+  bool Start() {
+    return sink_->Start(len_, &subc_);
+  }
+
+  bool End() {
+    return sink_->End();
+  }
+
+  // Puts a region of the given buffer [start, end) into the given sink (which
+  // probably represents a parser.  Can gracefully handle the case where the
+  // parser returns a "parsed" length that is less or greater than the input
+  // buffer length, and tracks the overall parse offset in *ofs.
+  //
+  // Pass verbose=true to print detailed diagnostics to stderr.
+  bool ParseBuffer(int bytes) {
+    if (bytes < 0) {
+      bytes = len_ - ofs_;
+    }
+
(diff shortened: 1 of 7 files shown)


tests/test_util.h
```diff
@@ -8,59 +8,90 @@
...
-// probably represents a parser.  Can gracefully handle the case where the
-// parser returns a "parsed" length that is less or greater than the input
-// buffer length, and tracks the overall parse offset in *ofs.
-//
-// Pass verbose=true to print detailed diagnostics to stderr.
-bool parse_buffer(upb::BytesSink* sink, void* subc, const char* buf,
-                  size_t start, size_t end, size_t* ofs,
-                  upb::Status* status, bool verbose) {
-  start = UPB_MAX(start, *ofs);
+class VerboseParserEnvironment {
```
scwhittle: add a comment

tests/test_util.h
```diff
@@ -8,59 +8,90 @@
...
+  bool End() {
+    return sink_->End();
+  }
+
+  // Puts a region of the given buffer [start, end) into the given sink (which
+  // probably represents a parser.  Can gracefully handle the case where the
+  // parser returns a "parsed" length that is less or greater than the input
+  // buffer length, and tracks the overall parse offset in *ofs.
+  //
+  // Pass verbose=true to print detailed diagnostics to stderr.
```
scwhittle: verbose is not parameter to this method.

tests/test_util.h
```diff
@@ -69,17 +100,55 @@ bool parse_buffer(upb::BytesSink* sink, void* subc, const char* buf,
...
   }
-  return true;
-}
+
+  const upb::Status& status() { return status_; }
+
+  size_t ofs() { return ofs_; }
+  upb::Environment* env() { return &env_; }
+
+  bool SkippedWithNull() { return skip_until_ > 0; }
```
scwhittle: confusing method name to me, nothing has actually skipped with null yet if this is set.

upb/pb/decoder.c
```diff
@@ -917,27 +965,6 @@ void upb_pbdecoder_reset(upb_pbdecoder *d) {
   d->residual_end = d->residual;
```
scwhittle: reset skip

tests/test_util.h
haberman: It does mean that something has been skipped with NULL or that the very next call to ParseBuffer() will skip with NULL (but it's intended to be used after End()). If you have a better name I'm all ears. :)

upb/pb/decoder.c
haberman: d->skip is set two lines above -- do you mean something else?

scwhittle: oops, that comment was supposed to be in upb_decoder_reset

tests/test_util.h
scwhittle: currently it is a confusing name if you call it before ParseBuffer since it hasn't skipped but will skip
This matches the name better to me, and enforces that the skipping was actually performed.
bool SkippedWithNull() { return skip_until_ > 0 && skip_until < ofs; }

upb/pb/decoder.c
haberman: So I reset skip in start(), which is in line with where I reset other things of its type. I haven't quite figured out yet what should be reset in reset() vs start() yet. I don't know what the semantics should be around: should a decoder be able to be used multiple times without a reset()? Should reset() just set some bit that allows for a start() when you weren't already in an initial state? Still figuring all this out. But the current code should work.

tests/test_util.h
haberman: That's not quite right either -- if there are multiple skips and we're just before a second one it will return false. I added an explicit flag and used it for the method. :)

tests/test_util.h
haberman: Done.

tests/test_util.h
haberman: Done.

haberman:
PTAL, tests are passing now.

scwhittle:
LGTM

PR #35, opened 2015-07-30
Switch to Travis's container-based infrastructure.

haberman:


travis.sh
```diff
@@ -20,30 +20,18 @@ barejit_script() {
 
 # Build with Google protobuf support and tests (with JIT).
 withprotobuf_install() {
-  sudo apt-get update -qq
-  sudo apt-get install protobuf-compiler libprotobuf-dev
+  :
```
scwhittle: are these empty install tuples still necessary?

otherwise LGTM

Issue #41, opened 2015-10-05
make testlua fails


bgmerrell:
Trying to run the lua bindings tests with Linux x86_64 (Ubuntu) and Lua 5.1, I get:

```
bgmerrell@oberon:~/code/upb$ make testlua USER_CPPFLAGS=`pkg-config lua5.1 --cflags`
CC -fPIC upb/pb/compile_decoder.c
CC -fPIC upb/pb/decoder.c
CC -fPIC upb/pb/encoder.c
CC -fPIC upb/pb/glue.c
CC -fPIC upb/pb/textprinter.c
CC -fPIC upb/pb/varint.c
AR lib/libupb.pb_pic.a
CC -fPIC upb/descriptor/descriptor.upb.c
CC -fPIC upb/descriptor/reader.c
AR lib/libupb.descriptor_pic.a
CC -fPIC upb/def.c
CC -fPIC upb/env.c
CC -fPIC upb/handlers.c
CC -fPIC upb/refcounted.c
CC -fPIC upb/shim/shim.c
CC -fPIC upb/symtab.c
CC -fPIC upb/table.c
CC -fPIC upb/upb.c
AR lib/libupb_pic.a
CC upb/bindings/lua/upb.c
CC upb/bindings/lua/upb/pb.c
LUA tests/bindings/lua/test_upb.lua
    ...............

406 Assertions checked.

Testsuite finished (15 passed, 0 failed, 0 errors).
LUA tests/bindings/lua/test_upb.pb.lua
lua: error loading module 'upb.pb_c' from file 'upb/bindings/lua/upb/pb_c.so':
        upb/bindings/lua/upb/pb_c.so: undefined symbol: lupb_openlib
stack traceback:
        [C]: ?
        [C]: in function 'require'
        upb/bindings/lua/upb/pb.lua:3: in main chunk
        [C]: in function 'require'
        tests/bindings/lua/test_upb.pb.lua:4: in main chunk
        [C]: ?
Makefile:444: recipe for target 'testlua' failed
make: *** [testlua] Error 1
```

I see the lupb_openlib symbol in the upb/bindings/lua/upb_c.so and upb/bindings/lua/upb/pb_c.so shared objects.

haberman:
[...]

PR #42, opened 2015-10-06
Fix lua bindings linking for Linux

bgmerrell:
This fixes an issue (#41) in Linux where lua could not lode the upd.pb_c
module due to an undefined symbol.  This update resolves the issue by
linking pb_c.so with upb_c.so that defines the symbol.

I've tested this on both Linux amd64 (Ubuntu, gcc) and Mac OS X Yosemite (LLVM).

I have signed the Google CLA previously.


bgmerrell:
@haberman Anything else needed for this?

haberman:
I apologize for not replying to this at the time. The Lua bindings have always been sort of experimental, and the build system has completely changed since you sent this (I've abandoned "make").

PR #43, opened 2015-10-15
Add flag to MessageDef for whether fields have presence.

haberman:
As a step towards proto2 (descriptor.proto) support.

Review to @tbetbetbe.


haberman:
Sounds like @tbetbetbe is on vacation, @scwhittle want to take a look?

Context is: we want to support proto2/proto3 simultaneously, which means having a flag in each message indicating whether it has proto2 presence semantics or proto3.

upb/def.c
```diff
@@ -1213,6 +1213,16 @@ bool upb_fielddef_ismap(const upb_fielddef *f) {
          upb_msgdef_mapentry(upb_fielddef_msgsubdef(f));
 }
 
+bool upb_fielddef_haspresence(const upb_fielddef *f) {
+  if (upb_fielddef_isseq(f)) return false;
+  if (upb_fielddef_issubmsg(f)) return true;
+
+  /* Primitive field: return true unless there is a message that specifies
+   * presence should not exist. */
+  if (f->msg_is_symbolic || !f->msg.def) return true;
```
scwhittle: would it be better to assert that the msg is not symbolic?
Returning true unconditionally seems incorrect

and a reminder, when would the message def be null?

upb/descriptor/reader.c
```diff
@@ -203,6 +209,28 @@ static size_t file_onpackage(void *closure, const void *hd, const char *buf,
...
 
+static size_t file_onsyntax(void *closure, const void *hd, const char *buf,
+                          size_t n, const upb_bufhandle *handle) {
+  upb_descreader *r = closure;
+  UPB_UNUSED(hd);
+  UPB_UNUSED(handle);
+  /* XXX: see comment at the top of the file. */
+  if (n == strlen("proto3") && memcmp(buf, "proto3", strlen("proto3")) == 0) {
+    uint32_t i;
+    /* Cover messages created previously. */
```
scwhittle: Cover doesn't seem like the best verb. How about
Update messages created before the syntax was read.

Update comment below too

upb/descriptor/reader.c
```diff
@@ -203,6 +209,28 @@ static size_t file_onpackage(void *closure, const void *hd, const char *buf,
   return n;
 }
 
+static size_t file_onsyntax(void *closure, const void *hd, const char *buf,
+                          size_t n, const upb_bufhandle *handle) {
+  upb_descreader *r = closure;
+  UPB_UNUSED(hd);
+  UPB_UNUSED(handle);
+  /* XXX: see comment at the top of the file. */
+  if (n == strlen("proto3") && memcmp(buf, "proto3", strlen("proto3")) == 0) {
```
scwhittle: verify syntax is either proto2 or proto3?

upb/structdefs.int.h
```diff
@@ -114,6 +114,11 @@ struct upb_msgdef {
    * descriptor.upb.c. */
   bool map_entry;
 
+  /* Do primitive values in this message have explicit presence or not?
+   * TODO: set this flag properly for static descriptors; regenerate
```
scwhittle: rm? seems this might be done already

upb/def.c
haberman: It would only be NULL before the field has been added to a message.

To me this is just implementing a sane default: presence defaults to true, but is false in cases where it is known to be false.

upb/structdefs.int.h
haberman: No, it hasn't been added to the code generator yet. The initialization macro always sets the flag to true right now.

upb/descriptor/reader.c
haberman: Done (once I upload the new PR).

upb/descriptor/reader.c
haberman: Done.

haberman:
PTAL.

PR #46, opened 2016-01-14
Changed JSON parser/printer to correctly camelCase names.

haberman:
This required adding support for json_name (and its camelCase transformation from the regular name) to upb_fielddef.

For now we are not adding a json_name -> fielddef* map to upb_msgdef. In many/most cases this will never be used. Instead we add the map to the JSON parser's data it creates up-front.


haberman:
Review to @scwhittle.

upb/def.c
```diff
@@ -721,6 +722,35 @@ const char *upb_fielddef_name(const upb_fielddef *f) {
   return upb_def_fullname(upb_fielddef_upcast(f));
 }
 
+bool upb_fielddef_getjsonname(const upb_fielddef *f, char *buf) {
```
scwhittle: why is this a method on fielddef instead of a different method taking any const char* name and camel casing?

If it has to be a method on fielddef, perhaps separate the transformation to a separate helper method that might be useful elsewhere for json

scwhittle: seems safer to pass in the size of buf and perform checks on that

upb/def.h
```diff
@@ -307,6 +307,12 @@ class upb::FieldDef {
   uint32_t number() const;   /* Returns 0 if uninitialized. */
   bool is_extension() const;
 
+  /* Get the JSON name for this field.  This will copy the JSON name into the
+   * given buffer, which must have size of at least "strlen(name()) + 1".
+   * The string will be NULL-terminated.  Returns false if uninitialized.
+   */
+  bool GetJsonName(char* buf) const;
```
scwhittle: C++ api seems like it would be easier to use if it took a string (and called reserve on it if current capacity is not enough)

upb/def.c
haberman: Camel casing the regular name is the *default* for the JSON name, but it can also explicitly be set to something else. See: https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto#L193

Separating out as a utility function could possibly be useful, but I'd rather late-bind on that.

upb/def.h
haberman: Done.

upb/def.c
haberman: Done on your second suggestion.

haberman:
Addressed comments, PTAL.

haberman:
Added asserts, PTAL.

scwhittle:
LGTM

PR #47, opened 2016-01-20
Optimized upb_inttable_compact(): it shrinks inttables more now.

haberman:
Review to @scwhittle.


upb/table.c
```diff
@@ -580,54 +584,49 @@ bool upb_inttable_removeptr(upb_inttable *t, const void *key, upb_value *val) {
...
-   * denotes an empty entry. */
-  arr_size = UPB_MAX(arr_size, 1);
+  assert(arr_count <= upb_inttable_count(t));
 
   {
     /* Insert all elements into new, perfectly-sized table. */
-    int hash_count = upb_inttable_count(t) - arr_count;
-    int hash_size = hash_count ? (hash_count / MAX_LOAD) + 1 : 0;
-    int hashsize_lg2 = log2ceil(hash_size);
+    size_t arr_size = max[size_lg2] + 1;
```
scwhittle: keep the comment that 0 must always be part of the array.

scwhittle:
LGTM with one comment

upb/table.c
haberman: I removed it because the constraint is commented and actually implemented elsewhere, see line 457 in upb_inttable_sizedinit().

The "+1" here is to accommodate the array being zero-based, so that the max key seen will actually fit at array[key].

haberman: Added a small explanatory comment for the +1. I'll merge this now.

PR #50, opened 2016-02-21
Handle oneof fields in DefBuilder

mbarbon:


upb/bindings/googlepb/bridge.cc
```diff
@@ -94,9 +94,23 @@ const MessageDef* DefBuilder::GetMaybeUnfrozenMessageDef(
     fields.push_back(d->field(i));
   }
 
+  // Oneof fields
+  for (int i = 0, maxi = d->oneof_decl_count(); i < maxi; ++i) {
```
haberman: Wrap in `#ifdef GOOGLE_PROTOBUF_HAS_ONEOF` so it can compile against protobuf < 2.6.0 (which introduced oneof support).

mbarbon:
Sorry about that. Added #ifdefs (not actually tested with an old protobuf release, though).

haberman:
Thanks for the contribution -- sorry I didn't merge it at the time. Since then I have removed this functionality: now to convert between protobuf descriptors and upb descriptors you just serialize and parse.

PR #52, opened 2016-03-17
Make OneofDef stop deriving from Def.

haberman:
It is clear now that oneofs are different form other
defs in one important way: they are not top-level
constructs that can stand on their own in a SymbolTable,
for example.  If you are iterating over a list of Defs
in a SymbolTable, there is no chance you will run into
a oneof.  To reflect this reality, OneofDef no longer
derives from Def, and the UPB_DEF_ONEOF is no longer
an enum value that needs to be handled.

### DIFF
--- a/upb/def.c
+++ b/upb/def.c
@@ -538,7 +538,7 @@ static void visitfield(const upb_refcounted *r, upb_refcounted_visit *visit,
     visit(r, upb_msgdef_upcast2(upb_fielddef_containingtype(f)), closure);
   }
   if (upb_fielddef_containingoneof(f)) {
-    visit(r, upb_oneofdef_upcast2(upb_fielddef_containingoneof(f)), closure);
+    visit(r, upb_oneofdef_upcast(upb_fielddef_containingoneof(f)), closure);
   }
   if (upb_fielddef_subdef(f)) {
     visit(r, upb_def_upcast(upb_fielddef_subdef(f)), closure);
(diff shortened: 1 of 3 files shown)


haberman:
@scwhittle want to take a look?

upb/def.c
```diff
@@ -1588,7 +1588,7 @@ static void freeoneof(upb_refcounted *r) {
   upb_oneofdef *o = (upb_oneofdef*)r;
   upb_strtable_uninit(&o->ntof);
   upb_inttable_uninit(&o->itof);
-  upb_def_uninit(upb_oneofdef_upcast_mutable(o));
+  free((void*)o->name);
```
scwhittle: is the cast to (void*) needed?

upb/def.c
```diff
@@ -1629,17 +1628,18 @@ upb_oneofdef *upb_oneofdef_dup(const upb_oneofdef *o, const void *owner) {
...
-                             upb_status *s) {
+bool upb_oneofdef_setname(upb_oneofdef *o, const char *name, upb_status *s) {
+  assert(!upb_oneofdef_isfrozen(o));
   if (upb_oneofdef_containingtype(o)) {
     upb_status_seterrmsg(s, "oneof already added to a message");
     return false;
   }
-  return upb_def_setfullname(upb_oneofdef_upcast_mutable(o), fullname, s);
+  if (!upb_isident(name, strlen(name), true, s)) return false;
+  free((void*)o->name);
```
scwhittle: ditto

scwhittle:
LGTM otherwise

upb/def.c
haberman: Yes because it's declared as const.

haberman:
Thanks! Merging.

PR #53, opened 2016-04-05
Added upb::FileDef, which represents the file defs are declared in.

haberman:
It is entirely optional: MessageDef/EnumDef can still exist
on their own.  But this can represent a def's file when it is
desirable to do so (eg. for code generators).

This approach will require that we change the way we handle
extensions.  But I think it will be a good change overall.

Specifically, we previously handled extensions by duplicating
the extended message and then adding the extension as a regular
field to the duplicated message.  This required also duplicating
any messages that could reach the extended message.

In the new world we will need a way of declaring and looking up
extensions separately from the message being extended.

This change also involves some notable changes to the generated
code:

- files are now called foo.upbdefs.h instead of foo.upb.h.
  This reflects the fact that we might possibly generate several
  different output files for a .proto file, and this one is just
  for defs.

- we no longer generate selectors in the .h file.

- the upbdefs.c no longer vends a SymbolTable.  Now it vends the
  individual messages (and possibly a FileDef later).  I think this
  will compose better once we can generate files where one
  generated files imports another.

We also make the descriptor reader vend a list of FileDefs now.
This is the best conceptual match for parsing a FileDescriptorSet.


haberman:
Review to @scwhittle. Sorry it's kind of a big one, but a good change I think.

tests/pb/test_encoder.cc
```diff
@@ -1,11 +1,31 @@
...
+  char *buf;
+  FILE *f = fopen(filename, "rb");
+  if(!f) return NULL;
+  if(fseek(f, 0, SEEK_END) != 0) goto error;
+  size = ftell(f);
+  if(size < 0) goto error;
+  if(fseek(f, 0, SEEK_SET) != 0) goto error;
+  buf = (char*)malloc(size + 1);
+  if(size && fread(buf, size, 1, f) != 1) goto error;
+  fclose(f);
```
scwhittle: buf[size] = '\0';

tests/test_def.c
```diff
@@ -38,15 +38,45 @@ static void test_noreftracking() {
...
+  char *buf;
+  FILE *f = fopen(filename, "rb");
+  if(!f) return NULL;
+  if(fseek(f, 0, SEEK_END) != 0) goto error;
+  size = ftell(f);
+  if(size < 0) goto error;
+  if(fseek(f, 0, SEEK_SET) != 0) goto error;
+  buf = malloc(size + 1);
+  if(size && fread(buf, size, 1, f) != 1) goto error;
+  fclose(f);
```
scwhittle: ditto, better yet can you share this utility function instead of reimplementing?

tools/dump_cinit.lua
```diff
@@ -552,18 +522,62 @@ local function dump_defs_c(symtab, basename, namespace, append)
...
   for _, def in ipairs(sorted) do
-    append(format, namespace, to_cident(def:full_name()), def:full_name())
+    append(format, to_cident(def:full_name()), def:full_name())
   end
 
   append("\n")
 end
 
+local function make_children_map(file)
+  -- Maps full_name -> children.
```
scwhittle: incorrect?  Appears you are mapping packages to their contained defs

upb/bindings/lua/upb.c
```diff
@@ -1056,6 +1064,128 @@ static const struct luaL_Reg lupb_enumdef_m[] = {
...
 
+/* lupb_filedef ***************************************************************/
+
+void lupb_filedef_pushwrapper(lua_State *L, const upb_filedef *f,
+                              const void *ref_donor) {
+  lupb_refcounted_pushwrapper(L, upb_filedef_upcast(f), LUPB_FILEDEF, ref_donor,
+                              sizeof(void *));
+}
+
+void lupb_filedef_pushnewrapper(lua_State *L, const upb_filedef *f,
```
scwhittle: should this be pushnewwrapper? ditto elsewhere

upb/bindings/lua/upb.c
```diff
@@ -1056,6 +1064,128 @@ static const struct luaL_Reg lupb_enumdef_m[] = {
...
+      return 1;
+    }
+  }
+
+  return 0;
+}
+
+static int lupb_filedef_defs(lua_State *L) {
+  lupb_filedef_check(L, 1);
+  luaL_checkinteger(L, 2);   /* Def type.  Could make this optional. */
```
scwhittle: would that be easy?  Seems like that would be convenient for all types

upb/def.c
```diff
@@ -73,6 +73,24 @@ upb_deftype_t upb_def_type(const upb_def *d) { return d->type; }
 
 const char *upb_def_fullname(const upb_def *d) { return d->fullname; }
 
+const char *upb_def_name(const upb_def *d) {
+  /* Return one past the last '.'. */
```
scwhittle: use strrchr?

upb/def.c
```diff
@@ -364,19 +387,24 @@ bool upb_def_freeze(upb_def *const* defs, int n, upb_status *s) {
    * limit. */
   maxdepth = UPB_MAX_MESSAGE_DEPTH * 2;
 
-  /* Validation all passed; freeze the defs. */
-  ret = upb_refcounted_freeze((upb_refcounted * const *)defs, n, s, maxdepth);
```
scwhittle: was this cast unsafe?  Otherwise it seems nice to have the type safety that the objs are upb_defs

upb/def.c
```diff
@@ -1493,7 +1529,8 @@ bool upb_msgdef_addoneof(upb_msgdef *m, upb_oneofdef *o, const void *ref_donor,
 
 void upb_msgdef_setprimitiveshavepresence(upb_msgdef *m, bool have_presence) {
   assert(!upb_msgdef_isfrozen(m));
-  m->primitives_have_presence = have_presence;
+  assert(!m->base.file);
+  m->syntax = have_presence ? UPB_SYNTAX_PROTO2 : UPB_SYNTAX_PROTO3;
```
scwhittle: how about changing this to be upb_msgdef_setsyntax?

upb/def.c
```diff
@@ -1757,3 +1794,179 @@ upb_fielddef *upb_oneof_iter_field(const upb_oneof_iter *iter) {
...
+    return upb_value_getconstptr(v);
+  } else {
+    return NULL;
+  }
+}
+
+bool upb_filedef_setname(upb_filedef *f, const char *name, upb_status *s) {
+  UPB_UNUSED(s);
+  free((void*)f->name);
+  f->name = upb_strdup(name);
```
scwhittle: do you want to handle strdup unable to allocate?

upb/def.c
```diff
@@ -1757,3 +1794,179 @@ upb_fielddef *upb_oneof_iter_field(const upb_oneof_iter *iter) {
...
+  return true;
+}
+
+bool upb_filedef_setpackage(upb_filedef *f, const char *package,
+                            upb_status *s) {
+  if (!upb_isident(package, strlen(package), true, s)) return false;
+  free((void*)f->package);
+  f->package = upb_strdup(package);
```
scwhittle: ditto

upb/def.h
```diff
@@ -1511,6 +1648,9 @@ inline reffed_ptr<MessageDef> MessageDef::New() {
 inline const char *MessageDef::full_name() const {
   return upb_msgdef_fullname(this);
 }
+inline const char *MessageDef::name() const {
+  return upb_msgdef_name(this);
+}
 inline bool MessageDef::set_full_name(const char* fullname, Status* s) {
```
scwhittle: std::string version?

upb/def.h
```diff
@@ -1698,6 +1838,9 @@ inline reffed_ptr<EnumDef> EnumDef::New() {
 inline const char* EnumDef::full_name() const {
   return upb_enumdef_fullname(this);
 }
+inline const char* EnumDef::name() const {
+  return upb_enumdef_name(this);
+}
 inline bool EnumDef::set_full_name(const char* fullname, Status* s) {
```
scwhittle: ditto

upb/def.h
```diff
@@ -1828,6 +1968,54 @@ inline bool OneofDef::const_iterator::operator!=(
...
 
+inline reffed_ptr<FileDef> FileDef::New() {
+  upb_filedef *f = upb_filedef_new(&f);
+  return reffed_ptr<FileDef>(f, &f);
+}
+
+inline const char* FileDef::name() const {
+  return upb_filedef_name(this);
+}
+inline bool FileDef::set_name(const char* name, Status* s) {
```
scwhittle: ditto for these

upb/symtab.c
```diff
@@ -393,11 +408,35 @@ err: {
...
+}
+
+bool upb_symtab_addfile(upb_symtab *s, upb_filedef *file, upb_status *status) {
+  size_t n;
+  size_t i;
+  upb_def **defs;
+  bool ret;
+
+  n = upb_filedef_defcount(file);
+  defs = malloc(sizeof(*defs) * n);
```
scwhittle: handle malloc error

tests/pb/test_encoder.cc
haberman: Done.

tests/test_def.c
haberman: Done.

tools/dump_cinit.lua
haberman: Both are true, key is a prefix (foo.bar.Baz) which can be either a package or a message with nested defs. Expanded slightly to reflect this.

upb/bindings/lua/upb.c
haberman: I think I was trying to increase readability by removing the repeated letter, a la getty: https://en.wikipedia.org/wiki/Getty_(Unix)

How does it strike you as a reader?

upb/bindings/lua/upb.c
haberman: It would be easy, I'm still trying to decide whether it's desirable. Maybe the explicitness of file:defs(upb.DEF_ANY) is good? Weighing advice like: https://google.github.io/styleguide/cppguide.html#Default_Arguments

What do you think?

upb/def.c

upb/def.c
haberman: I changed it because they no longer are always upb_defs -- we also freeze upb_filedefs (which are not upb_defs).

upb/def.c
haberman: I decided that having the msgdef field be called "syntax" makes the most sense, since this is how type.proto represents it (https://github.com/google/protobuf/blob/master/src/google/protobuf/type.proto#L58) and because a message's "syntax" affects more semantics than just field presence (for example, PROTO3 discards unknown fields by definition).

I think you're right though that this function can be removed. Done.

upb/def.c

upb/def.c

upb/def.h
haberman: It is right under this. :)  The review tool might have collapsed it.

upb/def.h
haberman: ditto

upb/symtab.c
haberman: Done.

upb/def.c
scwhittle: Now that I read the rest I see that the first add_n are all upb_defs but that the last freeze_n - add_n can be any refcounted object.
I think an intenral typed method taking defs to prepare for freezing would be cleaner:

 upb_freeze would be:
upb_def_prepare_for_freeze(defs, n);
upb_refcounted_freeze(defs, n);

and symtab_add would be:
upb_def_prepare_for_freeze(defs, n)
upb_refcounted_freeze(refcount_objs /*defs and file*/, freeze_n);

upb/def.h
haberman: You got me on this one and OneofDef. :) Fixed.

haberman:
Addressed all comments, PTAL.

scwhittle:
LGTM except for my suggestion for upb_def_freeze refactoring

haberman:
Refactoring suggestion implemented.

scwhittle:
LGTM

PR #54, opened 2016-04-10
Changed schema for JSON test to be defined in a .proto file.

haberman:
Before we had lots of code to build these schemas manually,
but this was verbose and made it difficult to add to the
schema easily.  Now we can just write a .proto file and
adding fields is easy.

To avoid making the tests depend on upbc (and thus Lua)
we check in the generated schema.


haberman:
Review to @scwhittle.

haberman:
Tests are green, want to take a look at this one also?

haberman:
@scwhittle.

scwhittle:
LGTM

PR #55, opened 2016-04-14
Added migration flag for users using old JSON format.

haberman:
This is intended for Ruby users who may currently be using JSON format. The Ruby JSON format was written before we realized that proto3 JSON would be camelCasing field names, so it always used original field names. The most recent upb code fixes this problem, but this could cause problems for any Ruby users who are currently using JSON.

These Ruby users should migrate, but it may be difficult to do all at once. So we are providing these flags for now to make the migration easier.

In the longer term, if Ruby users need compatibility with existing payloads, they should use the `json_name` option in their `.proto` files to specify the old names. Unfortunately proto3 Ruby doesn't support `json_name` yet. We will add support for it before we remove these flags.

### DIFF
--- a/upb/json/printer.c
+++ b/upb/json/printer.c
@@ -47,12 +47,19 @@ void freestrpc(void *ptr) {
 strpc *newstrpc(upb_handlers *h, const upb_fielddef *f) {
   /* TODO(haberman): handle malloc failure. */
   strpc *ret = malloc(sizeof(*ret));
-  size_t len;
-  ret->len = upb_fielddef_getjsonname(f, NULL, 0);
-  ret->ptr = malloc(ret->len);
-  len = upb_fielddef_getjsonname(f, ret->ptr, ret->len);
-  UPB_ASSERT_VAR(len, len == ret->len);
-  ret->len--;  /* NULL */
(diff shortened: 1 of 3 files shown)


haberman:
Review to @scwhittle.

upb/json/printer.c
```diff
@@ -47,12 +47,19 @@ void freestrpc(void *ptr) {
 strpc *newstrpc(upb_handlers *h, const upb_fielddef *f) {
   /* TODO(haberman): handle malloc failure. */
   strpc *ret = malloc(sizeof(*ret));
-  size_t len;
-  ret->len = upb_fielddef_getjsonname(f, NULL, 0);
-  ret->ptr = malloc(ret->len);
-  len = upb_fielddef_getjsonname(f, ret->ptr, ret->len);
-  UPB_ASSERT_VAR(len, len == ret->len);
-  ret->len--;  /* NULL */
+  if (getenv("UPB_JSON_WRITE_LEGACY_FIELD_NAMES")) {
```
scwhittle: Is this a performance concern? You could cache somewhere instead of looking up the env variable repeatedly

haberman: This code runs at handler creation time, not at print time.

scwhittle:
LGTM

PR #56, opened 2016-04-14
JSON parser: always accept both name variants, flag controls which we generate.

haberman:
My previous commit was based on wrong information about the
proto3 JSON spec.  It turns out we need to accept both field
name formats all the time, and a runtime flag should control
which we generate.


haberman:
Review to @scwhittle.

haberman:
btw I'm open to naming suggestions on the parameter.

scwhittle:
LGTM

PR #58, opened 2016-04-15
Split upb::Arena/upb::Allocator from upb::Environment.

haberman:
This will allow arenas and allocators to be used
independently of environments, which will be important
for an upcoming change (a message representation).
Overall this design feels cleaner that the previous
Environment/SeededAllocator design.

As part of this change, moved all allocations in upb
to use a global allocator instead of hard-coding
malloc/free.  This will allow injecting OOM faults
for more robust testing.

One place that doesn't use the global allocator is
the tracked ref code.  Instead of its previous approach
of CHECK_OOM() after every malloc() or table insert, it
simply uses an allocator that does this automatically.

I moved Allocator/Arena/Environment into upb.h.
This seems principled since these are the only types
in upb whose size is directly exposed to users, since
they form the basis of memory allocation strategy.


haberman:
Review to @scwhittle. I think you're going to like this!

haberman:
Tests are green now.

upb/def.c
```diff
@@ -89,9 +89,18 @@ const char *upb_def_name(const upb_def *d) {
 
 bool upb_def_setfullname(upb_def *def, const char *fullname, upb_status *s) {
   assert(!upb_def_isfrozen(def));
-  if (!upb_isident(fullname, strlen(fullname), true, s)) return false;
-  free((void*)def->fullname);
-  def->fullname = upb_strdup(fullname);
+  if (!upb_isident(fullname, strlen(fullname), true, s)) {
+    return false;
```
scwhittle: errmsg?

upb/def.c
```diff
@@ -1249,15 +1286,23 @@ bool upb_fielddef_setenumsubdef(upb_fielddef *f, const upb_enumdef *subdef,
...
+  char *name_copy;
   assert(!upb_fielddef_isfrozen(f));
   if (!upb_fielddef_hassubdef(f)) {
     upb_status_seterrmsg(s, "field type does not accept a subdef");
     return false;
   }
+
+  name_copy = upb_gstrdup(name);
+  if (!name_copy) {
+    upb_status_seterrmsg(s, "Out of memory.");
```
scwhittle: Maybe a helper method for this? The same string is popping up a lot

upb/descriptor/reader.c
```diff
@@ -75,9 +75,9 @@ static char *upb_strndup(const char *buf, size_t n) {
  * Caller owns a ref on the returned string. */
 static char *upb_join(const char *base, const char *name) {
   if (!base || strlen(base) == 0) {
-    return upb_strdup(name);
+    return upb_gstrdup(name);
   } else {
-    char *ret = malloc(strlen(base) + strlen(name) + 2);
+    char *ret = upb_gmalloc(strlen(base) + strlen(name) + 2);
```
scwhittle: handle oom

upb/handlers.c
```diff
@@ -6,11 +6,19 @@
...
-#include <stdlib.h>
 #include <string.h>
 
 #include "upb/sink.h"
 
+static void *upb_calloc(size_t size) {
+  void *mem = upb_gmalloc(size);
+  if (mem) {
+    memset(mem, 0, size);
+    return mem;
```
scwhittle: shorter as
if (mem) memset(...)
return mem;

upb/table.c
```diff
@@ -454,42 +476,45 @@ static void check(upb_inttable *t) {
...
-  free(mutable_array(t));
+void upb_inttable_uninit2(upb_inttable *t, upb_alloc *a) {
+  uninit(&t->t, a);
+  upb_free(a, mutable_array(t));
 }
 
-bool upb_inttable_insert(upb_inttable *t, uintptr_t key, upb_value val) {
+bool upb_inttable_insert2(upb_inttable *t, uintptr_t key, upb_value val,
+                          upb_alloc *a) {
   /* XXX: Table can't store value (uint64_t)-1.  Need to somehow statically
```
scwhittle: add an assert for this at least?

upb/table.int.h
```diff
@@ -63,10 +63,14 @@ typedef struct {
 #endif
 
 /* Like strdup(), which isn't always available since it's not ANSI C. */
-char *upb_strdup(const char *s);
+char *upb_strdup(const char *s, upb_alloc *a);
```
scwhittle: this seems like an odd file for this, it's used by other things than just table now, right?

upb/upb.h
```diff
@@ -267,179 +282,100 @@ class PointerBase2 : public PointerBase<T, Base> {
...
-   * object.  Instead we provide this: */
-  T* ReleaseTo(const void* new_owner) {
-    T* ret = NULL;
-    ptr_->DonateRef(this, new_owner);
-    std::swap(ret, ptr_);
-    return ret;
-  }
+/* Errors raised by upb that we want to be able to detect programmatically. */
+typedef enum {
+  UPB_NOMEM   /* Can't reuse ENOMEM because it is POSIX, not ISO C. */
```
scwhittle: consider using the canonical codes instead, grpc uses them for example
https://godoc.org/google.golang.org/grpc/codes

upb/upb.h
```diff
@@ -458,56 +394,309 @@ struct upb_status {
...
+ * compatibility even if the size of the objects change.  And this allows the
+ * structure definitions to be in the .c files instead of the .h files, making
+ * the .h files smaller and more readable.
+ *
+ * We might want to consider renaming this to "Pipeline" if/when the concept of
+ * a pipeline element becomes more formalized. */
+UPB_DECLARE_TYPE(upb::Environment, upb_env)
+
+/* A function that receives an error report from an encoder or decoder.  The
+ * callback can return true to request that the error should be recovered, but
```
scwhittle: do you mean retried? recovered is confusing to me

upb/def.c
haberman: upb_isident() takes care of that.

upb/upb.h
haberman: Thanks for the reference, but it doesn't seem like a good fit. Most of the error codes wouldn't apply to upb (unauthenticated, deadline exceeded, permission denied, unavailable, data loss). But they also don't differentiate between out of memory and exceeded resource limit, which to me are very different cases.

upb/table.int.h
haberman: Nope, just table.c right now. reader.c has `upb_strndup()`. If either spreads beyond one file I'll move it to `upb.h`.

upb/table.c
haberman: I think the comment is obsolete. I removed it and added a test that this case works.

In the process I noticed that test_table.cc was kind of stale and I added some C++ template helpers. I got a little carried away with this -- let me know if you think it's a net improvement. If not I'll revert and just have the added int64 test. I think the table should probably be tested more exhaustively and I thought the templated convenience classes might help with that later on.

upb/descriptor/reader.c
haberman: Done.

upb/handlers.c
haberman: Done.

upb/table.c
haberman: Whoops, I was wrong about that. The comment still applies. We already had an assert for this case actually. I made the comment more specific. I should find a way to fix this.

haberman:
Addressed comments, PTAL.

upb/descriptor/reader.c
```diff
@@ -87,14 +90,20 @@ static char *upb_join(const char *base, const char *name) {
...
   size_t i;
   for (i = start; i < upb_filedef_defcount(f); i++) {
     upb_def *def = upb_filedef_mutabledef(f, i);
     char *name = upb_join(str, upb_def_fullname(def));
+    if (!name) {
+      /* Need better logic here; at this point we've qualified some names but
+       * not others. */
+      return false;
+    }
     upb_def_setfullname(def, name, NULL);
```
scwhittle: this may fail with oom too

upb/upb.h
```diff
@@ -388,14 +375,35 @@ struct upb_status {
   int code_;
   upb_errorspace *error_space_;
 
+  const char *file_;
```
scwhittle: are these used?

scwhittle:
LGTM

upb/descriptor/reader.c
haberman: There are probably lots of places like this, and we don't have a great recovery story here. I'm going to punt on being super thorough with this until we're actually testing for it anyway by injecting malloc failures.

haberman:
Thanks! Merging.

PR #60, opened 2016-04-20
Put oneofs in the same table as fields.

haberman:
Oneofs and fields are not allowed to have names that conflict,
so we might as well put them all in the same table.  This also
allows an efficient operation that looks for both fields and
oneofs in a single lookup.

Added support for OneofDef to Lua to allow testing of this.

### DIFF
--- a/tests/bindings/lua/test_upb.lua
+++ b/tests/bindings/lua/test_upb.lua
@@ -16,6 +16,19 @@ function iter_to_array(iter)
   return arr
 end
 
+function test_msgdef()
+  local o = upb.OneofDef{name = "field1", fields = {
+    upb.FieldDef{name = "field2", number = 1, type = upb.TYPE_INT32},
+  }}
+
+  local f = upb.FieldDef{name = "field3", number = 2, type = upb.TYPE_INT32}
+
+  local m = upb.MessageDef{fields = {o, f}}
+
+  assert_equal(f, m:lookup_name("field3"))
+  assert_equal(o, m:lookup_name("field1"))
+end
+
 function test_fielddef()
   local f = upb.FieldDef()
   assert_false(f:is_frozen())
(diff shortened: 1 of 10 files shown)


haberman:
Review to @scwhittle.

tests/bindings/lua/test_upb.lua
```diff
@@ -16,6 +16,19 @@ function iter_to_array(iter)
...
+  local o = upb.OneofDef{name = "field1", fields = {
+    upb.FieldDef{name = "field2", number = 1, type = upb.TYPE_INT32},
+  }}
+
+  local f = upb.FieldDef{name = "field3", number = 2, type = upb.TYPE_INT32}
+
+  local m = upb.MessageDef{fields = {o, f}}
+
+  assert_equal(f, m:lookup_name("field3"))
+  assert_equal(o, m:lookup_name("field1"))
```
scwhittle: lookup field2 also?

upb/def.c
```diff
@@ -1484,7 +1508,7 @@ static bool check_field_add(const upb_msgdef *m, const upb_fielddef *f,
   } else if (upb_fielddef_name(f) == NULL || upb_fielddef_number(f) == 0) {
     upb_status_seterrmsg(s, "field name or number were not set");
     return false;
-  } else if (upb_msgdef_ntofz(m, upb_fielddef_name(f)) ||
+  } else if (upb_strtable_lookup(&m->ntof, upb_fielddef_name(f), NULL) ||
              upb_msgdef_itof(m, upb_fielddef_number(f))) {
     upb_status_seterrmsg(s, "duplicate field name or number for field");
```
scwhittle: same message as below, mention oneofs

upb/def.c
```diff
@@ -1589,23 +1613,46 @@ const upb_fielddef *upb_msgdef_itof(const upb_msgdef *m, uint32_t i) {
...
+bool upb_msgdef_lookupname(const upb_msgdef *m, const char *name, size_t len,
+                           const upb_fielddef **f, const upb_oneofdef **o) {
+  upb_value val;
+
+  if (!upb_strtable_lookup2(&m->ntof, name, len, &val)) {
+    return false;
+  }
+
+  *o = upb_trygetoneof(upb_value_getptr(val));
+  *f = upb_trygetfield(upb_value_getptr(val));
```
scwhittle: assert not both are set

scwhittle:
LGTM with minor comments

upb/def.c

upb/def.c
haberman: Done (I made it slightly stronger and asserted that exactly one is set).

PR #62, opened 2016-05-12
Added UPB_ASSERT() that helps avoid unused var warnings.

haberman:


haberman:
Review to @acozzette.

haberman:
The change looks big, but it's mostly very mechanical. The part to look at more closely is upb/upb.h. There are also a couple places where we were able to remove `UPB_ASSERT_VAR` in favor of the simpler `UPB_ASSERT`, and where we could remove `UPB_UNUSED()` declarations.

upb/upb.h
```diff
@@ -196,10 +196,17 @@ template <int N> class InlinedEnvironment;
...
 #define UPB_UNUSED(var) (void)var
 
-/* For asserting something about a variable when the variable is not used for
- * anything else.  This prevents "unused variable" warnings when compiling in
- * debug mode. */
-#define UPB_ASSERT_VAR(var, predicate) UPB_UNUSED(var); assert(predicate)
+/* UPB_ASSERT(): in release mode, we use the expression without letting it be
+ * evaluated.  This prevents "unused variable" warnings. */
+#ifdef NDEBUG
+#define UPB_ASSERT(expr) if (false && (expr)) {}
```
acozzette: Nit: I think this macro could theoretically cause problems if you use it inside an if statement without braces. E.g. I think the code below would do something unexpected:

```
if (whatever)
  UPB_ASSERT(foo)
else
  // ...
```

In this case I think the else statement will always be executed because it will associate itself with the if statement inside the macro. Maybe this would be one way to fix it:

`#define UPB_ASSERT(expr) do {} while (false && (expr))`

If I'm not mistaken this also has the advantage of requiring a semicolon at the end when the macro is used, so then there's no ambiguity about whether you need to put a semicolon or not.

upb/upb.h
```diff
@@ -196,10 +196,17 @@ template <int N> class InlinedEnvironment;
...
+#else
+#define UPB_ASSERT(expr) assert(expr)
+#endif
+
+/* UPB_ASSERT_DEBUGVAR(): assert that uses functions or variables that do not
```
acozzette: Maybe I'm confused but shouldn't this say functions or variables that _only_ exist in debug mode?

upb/upb.h
acozzette: Actually, now that I think of it, I thought in C you could just do `(void) var;` to reference a variable without using it; is there any reason that doesn't work here?

acozzette: Oh, never mind, I see you already have that above and it doesn't work well for arbitrary expressions.

haberman: `(void) var` evaluates the expression and throws away the result. It's important that `UPB_ASSERT()` in release mode *not* evaluate the expression. When I change this to use `UPB_UNUSED()` instead, it grows the binary in release mode (and presumably slows it down too).

haberman: The `do/while` suggestion is good -- thanks!

upb/upb.h
haberman: Good catch!

acozzette:
LGTM

PR #64, opened 2016-07-03
Mark auto-generated map entry fields for proto3

mbarbon:
Propagates the map_entry flag for proto3 map<> fields.


haberman:
Ditto on the PR, this functionality has been removed.

PR #73, opened 2016-11-13
Fix musl libc incompatibility

apstndb:
This library is not compatible with musl libc because the `__va_copy` function is glibc internal function.
I found the problem when I am trying to use protobuf in ruby. ( See https://github.com/google/protobuf/issues/2335#issuecomment-259887751
 )

The incompatibility can be fixed by using standard `va_copy` function whenever it is being built in C99( and C99-compatible C++11 ). 

It is testable using the following Dockerfile.

```
FROM alpine

RUN apk update
RUN apk add build-base

WORKDIR /root

COPY . ./

RUN make CFLAGS='-std=c99'
RUN make CFLAGS='-std=c99' CXXFLAGS='-std=c++11' tests test
```

### DIFF
--- a/tests/pb/test_decoder.cc
+++ b/tests/pb/test_decoder.cc
@@ -111,7 +111,7 @@ using std::string;
 
 void vappendf(string* str, const char *format, va_list args) {
   va_list copy;
-  __va_copy(copy, args);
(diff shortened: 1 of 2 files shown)


apstndb:
I sign it!

PR #74, opened 2016-11-29
Added upb_msg and Lua bindings for using it.

haberman:
upb_msg is a major step forward for upb. We will finally have a standard message representation, but one that interoperates with different memory management schemes (arena allocation, unique ownership, etc). This will also pave the way for representing custom options properly.

This PR is pretty big, but it's not quite as big as it looks, because several things were shuffled around.

In particular, the main`upb.c` in the Lua bindings was split into 3: `upb.c`, `msg.c`, and `def.c`. So basically none of the code in `def.c` is actually new, even though GitHub is showing it all as added in the PR.

Also, `upb/shim/*` functionality was moved into `upb/msg.c`.

There is some dead code here: functionality that I implemented in `upb_msg` but is not tested yet (like maps, extensions, repeated string fields, etc). I didn't want to make this CL grow even more, so I decided to stop when I had parity with what the Lua implementation could do before this change. Follow-up CLs will flesh out the tests to exercise all these cases, and will fix the inevitable bugs.

Review to @scwhittle.


haberman:
Tests are all passing now! Should be ready for review @scwhittle.

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
+
+#include "upb/msg.h"
+
+static bool is_power_of_two(size_t val) {
```
scwhittle: consider extracting these utility methods to somewhere reusable

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+         type == UPB_TYPE_UINT64 || type == UPB_TYPE_STRING;
+}
+
+void *upb_array_pack(const upb_array *arr, void *p, size_t *ofs, size_t size);
+void *upb_map_pack(const upb_map *map, void *p, size_t *ofs, size_t size);
+
+
+/** upb_msgval ****************************************************************/
+
+#define upb_alignof(t) offsetof(struct { char c; t x; }, x)
```
scwhittle: ditto

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+  size_t extdict_offset;
+  void *default_msg;
+  uint32_t *offsets;
+  uint32_t *case_offsets;
+  uint32_t *hasbits;
+  bool has_extdict;
+  uint8_t align;
+};
+
+static void upb_msg_check(const upb_msglayout *l, const upb_fielddef *f) {
```
scwhittle: upb_msglayout_checkfield?

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+  return l->hasbits[upb_fielddef_index(f)];
+}
+
+static upb_msglayout *upb_msglayout_new(const upb_msgdef *m) {
+  upb_msg_field_iter it;
+  upb_msg_oneof_iter oit;
+  upb_msglayout *l;
+  size_t hasbit;
+
+  l = upb_gmalloc(sizeof(*l));
```
scwhittle: seems like you could allocate a single block with arrays interned instead of 4 allocs

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+      l->offsets[upb_fielddef_index(f)] = val_offset;
+    }
+  }
+
+  /* Size of the entire structure should be a multiple of its greatest
+   * alignment. */
+  l->size = align_up(l->size, l->align);
+
+  if (upb_msgdef_syntax(m) == UPB_SYNTAX_PROTO2 && l->size) {
+    /* Allocate default message and set default values in it. */
```
scwhittle: have a separate method for this

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+                                              const upb_msgdef *m) {
+  upb_value v;
+  UPB_ASSERT(upb_symtab_lookupmsg(f->symtab, upb_msgdef_fullname(m)) == m);
+  UPB_ASSERT(!upb_msgdef_mapentry(m));
+
+  if (upb_inttable_lookupptr(&f->layouts, m, &v)) {
+    UPB_ASSERT(upb_value_getptr(v));
+    return upb_value_getptr(v);
+  } else {
+    /* This method should be made thread-safe so the "const" is accurate. */
```
scwhittle: seems safer to make the method take a non-const factory until this is fixed with a todo at the definition.

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+  uint32_t ofs = (uintptr_t)hd;
+  /* We pass NULL here because we know we can get away with it. */
+  upb_alloc *alloc = upb_msg_alloc(msg, NULL);
+  upb_msgval val;
+  UPB_UNUSED(size_hint);
+
+  val = upb_msgval_read(msg, ofs, upb_msgval_sizeof(UPB_TYPE_STRING));
+
+  upb_free(alloc, (void*)val.str.ptr);
+  val.str.ptr = NULL;
```
scwhittle: take advantage of size_hint?

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+    } else {
+      upb_msg_setscalarhandler(
+          h, f, offset, upb_msglayout_hasbit(layout, f));
+    }
+  }
+}
+
+const upb_handlers *upb_msgfactory_getmergehandlers(const upb_msgfactory *f,
+                                                    const upb_msgdef *m) {
+  /* This method should be made thread-safe so the "const" is accurate. */
```
scwhittle: ditto, seems safer to change parameter for now

upb/msg.c
```diff
@@ -0,0 +1,1053 @@
...
+  const char *key_str;
+  size_t key_len;
+  upb_value tabval = upb_toval(val);
+  upb_value removedtabval;
+  upb_alloc *a = map->alloc;
+
+  upb_map_tokey(map->key_type, &key, &key_str, &key_len);
+
+  if (upb_strtable_lookup2(&map->strtab, key_str, key_len, NULL)) {
+    upb_strtable_remove3(&map->strtab, key_str, key_len, &removedtabval, a);
```
scwhittle: this can end up doing three finds: lookup, remove and insert
consider adding an overwrite method that returns previous value?

upb/msg.h
```diff
@@ -0,0 +1,392 @@
...
+size_t upb_mapiter_sizeof();
+
+void upb_mapiter_begin(upb_mapiter *i, const upb_map *t);
+upb_mapiter *upb_mapiter_new(const upb_map *t, upb_alloc *a);
+void upb_mapiter_free(upb_mapiter *i, upb_alloc *a);
+void upb_mapiter_next(upb_mapiter *i);
+bool upb_mapiter_done(const upb_mapiter *i);
+
+/* For string keys, the value will be in upb_msgval_strkey(), *not*
+ * upb_msgval_str(). */
```
scwhittle: can this be avoided?

tests/bindings/lua/test_upb.lua
```diff
@@ -567,18 +568,19 @@ end
...
         }
       }
-    )
+    }
+
+    factory = upb.MessageFactory(symtab)
+    TestMessage = factory:get_message_class(symtab:lookup("TestMessage"))
+    msg = TestMessage()
 
     -- Defaults to nil
```
haberman: Done.

tests/bindings/lua/test_upb.lua
```diff
@@ -661,15 +664,17 @@ function test_msg_primitives()
...
       upb.FieldDef{name = "i32_array", number = 1, type = upb.TYPE_INT32,
                    label = upb.LABEL_REPEATED},
       }
     }
   }
 
-  msg = upb.Message(msgdef)
+  factory = upb.MessageFactory(symtab)
+  TestMessage = factory:get_message_class(symtab:lookup("TestMessage"))
+  msg = TestMessage()
```
haberman: Good point. I changed this one to take a name (only). Can extend it later to take a msgdef also if desired (can dispatch based on its type).

upb/msg.h
haberman: Whoops, that was an obsolete comment, removed.

upb/msg.c
haberman: I prefer to keep them local until/unless they are needed in more than one place.

upb/msg.c

upb/msg.c

upb/msg.c

upb/msg.c

upb/msg.c

upb/msg.c
haberman: Added TODO.

upb/msg.c
haberman: Cant easily do; size_hint isn't guaranteed, and we don't have enough space to store both the string's actual length and its allocated length.

size_hint isn't turning out to be very useful in practice...

haberman:
Addressed comments, PTAL.

upb/msg.c
```diff
@@ -193,27 +192,59 @@ static uint32_t upb_msglayout_hasbit(const upb_msglayout *l,
...
-
-  if (upb_msgdef_syntax(m) == UPB_SYNTAX_PROTO2) {
-    l->hasbits = upb_gmalloc(sizeof(*l->hasbits) * upb_msgdef_numfields(m));
-  }
-
-  l->offsets = upb_gmalloc(sizeof(*l->offsets) * upb_msgdef_numfields(m));
-  l->case_offsets =
-      upb_gmalloc(sizeof(*l->case_offsets) * upb_msgdef_numoneofs(m));
   l->align = 1;
+  l->offsets = (uint32_t*)CHARPTR_AT(l, sizeof(*l));
```
scwhittle: field_offsets?

upb/msg.c
```diff
@@ -361,13 +368,17 @@ void upb_msgfactory_free(upb_msgfactory *f) {
   upb_gfree(f);
 }
 
+const upb_symtab *upb_msgfactory_symtab(const upb_msgfactory *f) {
+  return f->symtab;
+}
+
 /* Requires:
  * - m is in upb_msgfactory_symtab(f)
```
scwhittle: update comment, maybe add todo about making it threadsafe and removing non-const method?

haberman: Removed comment since it was redundant from .h file. I added a TODO in the .h file.

upb/msg.c

PR #75, opened 2016-12-12
Basic serialization for upb_msg and Lua.

haberman:
Doesn't yet include strings, submessages, maps,
or repeated fields.


haberman:
Review to @scwhittle. This one is much more reasonably sized. :)

haberman:
Oh this PR also deletes a C file that was mistakenly added to the previous change (I accidentally had essentially a duplicate file under a different name).

upb/bindings/lua/upb/pb.c
```diff
@@ -32,20 +33,51 @@ static int lupb_pb_strtomessage(lua_State *L) {
...
   upb_env_uninit(&env);
-
   lupb_checkstatus(L, &status);
 
   /* References the arena at the top of the stack. */
   lupb_msg_pushref(L, lua_upvalueindex(3), msg);
   return 1;
 }
 
+static int lupb_pb_messagetostr(lua_State *L) {
```
scwhittle: lupb_pb_endodemsg?
It doesn't seem like it needs to be string specific depending on what the handlers do right?

upb/msg.c
```diff
@@ -472,6 +479,150 @@ const upb_handlers *upb_msgfactory_getmergehandlers(upb_msgfactory *f,
...
+static upb_selector_t getsel(const upb_fielddef *f, upb_handlertype_t type) {
+  upb_selector_t ret;
+  bool ok = upb_handlers_getselector(f, type, &ret);
+  UPB_ASSERT(ok);
+  return ret;
+}
+
+static bool upb_visitor_hasfield(const upb_msg *msg, const upb_fielddef *f,
+                                 const upb_msglayout *layout) {
+  if (upb_fielddef_isseq(f) || upb_fielddef_issubmsg(f)) {
```
scwhittle: is the submsg check necessary here?  is it not covered by the switch below?

upb/msg.c
```diff
@@ -472,6 +479,150 @@ const upb_handlers *upb_msgfactory_getmergehandlers(upb_msgfactory *f,
...
+static void upb_visitor_visitmsg2(const upb_msg *msg,
+                                  const upb_msglayout *layout, upb_sink *sink,
+                                  int depth) {
+  const upb_msgdef *md = upb_msglayout_msgdef(layout);
+  upb_msg_field_iter i;
+  upb_status status;
+
+  upb_sink_startmsg(sink);
+
+  /* Protect against cycles (possible because users may freely reassign message
```
scwhittle: surface this somehow?

upb/sink.c
```diff
@@ -0,0 +1,88 @@
...
+  UPB_UNUSED(hd);
+  UPB_UNUSED(size_hint);
+  sink->len = 0;
+  return sink;
+}
+
+static size_t upb_bufsink_string(void *_sink, const void *hd, const char *ptr,
+                                size_t len, const upb_bufhandle *handle) {
+  upb_bufsink *sink = _sink;
+  size_t new_size = sink->size;
```
scwhittle: make sure new_size is not zero, otherwise it will loop forever below

upb/bindings/lua/upb/pb.c
haberman: This specific function is specific to string. If later we want to encode to something else, we could either make a new specific function or factor the whole thing to be more general. But I'm not confident about what the general API would look like yet, so it seems better to stay specific.

upb/msg.c
haberman: Good call! Done.

upb/sink.c
haberman: It gets initialized to 32. I added an assert that it's not zero.

haberman:
Addressed comments, PTAL.

PR #76, opened 2017-01-21
Moved upb_symtab to def.h/def.c.

haberman:
This is in anticipation of removing refcounting and
making upb_symtab (soon to be upb_defpool) the unique
owner of all defs inside.


haberman:
Review to @scwhittle.

haberman:
Thanks! Not really -- this is mainly a prelude to more substantial changes. I wanted to separate the move so it was clear what was changing in the follow-up.

PR #77, opened 2017-01-23
Simplify and remove some code from upb::SymbolTable

haberman:
This makes `upb::SymbolTable` no longer refcounted. It also rips out the code to allow replacement of defs in a SymbolTable; this capability was not useful and was buggy. In the future we'll follow the model of the main protobuf implementation: messages can't be redefined.

We do keep some of the legacy "dup" code around because it's currently used for extensions. However we will eventually want to remove this also and handle extensions in a different way.

### DIFF
--- a/upb/def.c
+++ b/upb/def.c
@@ -2142,8 +2142,7 @@ bool upb_filedef_adddep(upb_filedef *f, const upb_filedef *dep) {
   }
 }
 
-static void upb_symtab_free(upb_refcounted *r) {
-  upb_symtab *s = (upb_symtab*)r;
+void upb_symtab_free(upb_symtab *s) {
   upb_strtable_iter i;
   upb_strtable_begin(&i, &s->symtab);
   for (; !upb_strtable_done(&i); upb_strtable_next(&i)) {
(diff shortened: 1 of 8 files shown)


upb/def.c
```diff
@@ -2372,12 +2236,15 @@ static bool symtab_add(upb_symtab *s, upb_def *const*defs, size_t n,
...
-       * need to donate the ref back to the caller. */
-      def->came_from_user = true;
+      if (upb_strtable_lookup(&s->symtab, fullname, NULL)) {
+        upb_status_seterrf(status, "Symtab already has a def named '%s'",
+                           fullname);
+        goto err;
+      }
       upb_def_donateref(def, ref_donor, s);
       if (!upb_strtable_insert(&addtab, fullname, upb_value_ptr(def)))
         goto oom_err;
```
scwhittle: in the error condition here it appears it won't undonate the ref of the current def

should you move the ref donation after adding successfully to addtab?

upb/def.c
```diff
@@ -2533,18 +2389,11 @@ static bool symtab_add(upb_symtab *s, upb_def *const*defs, size_t n,
 oom_err:
```
scwhittle: line 2379 above appears unnecessary now since the symtab should not contain anything with the same name

upb/def.c
```diff
@@ -2372,12 +2236,15 @@ static bool symtab_add(upb_symtab *s, upb_def *const*defs, size_t n,
...
-      def->came_from_user = true;
+      if (upb_strtable_lookup(&s->symtab, fullname, NULL)) {
+        upb_status_seterrf(status, "Symtab already has a def named '%s'",
+                           fullname);
+        goto err;
+      }
       upb_def_donateref(def, ref_donor, s);
       if (!upb_strtable_insert(&addtab, fullname, upb_value_ptr(def)))
         goto oom_err;
+      def->came_from_user = true;
```
scwhittle: remove this member? it appears unused now

upb/def.c
haberman: Good call, done. One benefit of the direction we're going is that the symtab will have an allocator, and we can use that to inject `malloc()` failure into all these paths.

upb/def.c
haberman: It was there for the old way of adding extensions. But I just ripped that out too, because I don't think it was actually being used. So extension support is removed until the new, more proto2-like method is implemented.

upb/def.c

PR #80, opened 2017-03-14
Add new file option php_class_prefix.

TeBoring:
This option will be prepended to all php generated classes. The PHP
runtime needs to know this option to figure out the class name for
specific message.


haberman:
LGTM.

PR #83, opened 2017-05-30
Add new file option php_namespace.

TeBoring:
Use this option to change the namespace of php generated classes.
Default is empty. When this option is empty, the package name will be
used for determining the namespace.


TeBoring:
Needed for https://github.com/google/protobuf/issues/3122

PR #87, opened 2017-07-02
Start migrating upb_msglayout to be suitable for generated code.

haberman:
This involves:

- remove upb_msglayout -> upb_msgfactory dependency.
- remove upb_msglayout -> upb_msgdef dependency.
- make upb_msglayout use a representation that can be
  statically initialized by generated code.

The goal here is that upb_msglayout becomes a kind of "descriptor
lite": it contains enough data to parser and serialize protobufs
and manipulate a upb_msg in memory, while being far smaller and
simpler than a full descriptor.  It also does not include field
names, which can be a benefit for applications that do not want
to leak field names.

Generated code can then create a upb_msglayout, and do most things
without ever needing to construct full descriptors/defs if they
don't want to.

### DIFF
--- a/upb/msg.c
+++ b/upb/msg.c
@@ -701,13 +701,15 @@ size_t upb_msg_sizeof(const upb_msglayout *l) {
 
 upb_msg *upb_msg_init(void *mem, const upb_msglayout *l, upb_alloc *a) {
   upb_msg *msg = VOIDPTR_AT(mem, upb_msg_internalsize(l));
+
+  /* Initialize normal members. */
   if (l->data.default_msg) {
     memcpy(msg, l->data.default_msg, l->data.size);
   } else {
     memset(msg, 0, l->data.size);
   }
 
-  UPB_ASSERT(!upb_msg_getinternal(msg)->alloc);
+  /* Initialize internal members. */
   upb_msg_getinternal(msg)->alloc = a;
 
   if (l->data.extendable) {
--- a/upb/msg.c
+++ b/upb/msg.c
@@ -24,7 +24,7 @@ bool upb_fieldtype_mapkeyok(upb_fieldtype_t type) {
 void *upb_array_pack(const upb_array *arr, void *p, size_t *ofs, size_t size);
 void *upb_map_pack(const upb_map *map, void *p, size_t *ofs, size_t size);
 
-#define CHARPTR_AT(msg, ofs) ((char*)msg + ofs)
+#define VOIDPTR_AT(msg, ofs) (void*)((char*)msg + ofs)
 #define ENCODE_MAX_NESTING 64
 #define CHECK_TRUE(x) if (!(x)) { return false; }
 
@@ -150,16 +150,8 @@ static upb_msgval upb_msgval_fromdefault(const upb_fielddef *f) {
 /** upb_msglayout *************************************************************/
 
 struct upb_msglayout {
-  upb_msgfactory *factory;
-  const upb_msgdef *msgdef;
-  size_t size;
-  size_t extdict_offset;
-  void *default_msg;
-  uint32_t *field_offsets;
-  uint32_t *case_offsets;
-  uint32_t *hasbits;
-  bool has_extdict;
-  uint8_t align;
+  const upb_msgdef *msgdef;  /* TODO(haberman): remove. */
+  struct upb_msglayout_msginit_v1 data;
 };
 
 static void upb_msg_checkfield(const upb_msglayout *l, const upb_fielddef *f) {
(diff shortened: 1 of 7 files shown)


upb/msg.c
```diff
@@ -232,22 +225,36 @@ static upb_msglayout *upb_msglayout_new(const upb_msgdef *m) {
...
   if (!l) return NULL;
 
   memset(l, 0, sizeof(*l));
 
-  l->msgdef = m;
-  l->align = 1;
-  l->field_offsets = (uint32_t*)CHARPTR_AT(l, sizeof(*l));
-  l->case_offsets = l->field_offsets + upb_msgdef_numfields(m);
-  l->hasbits = l->case_offsets + upb_msgdef_numoneofs(m);
+  /* TODO(haberman): check OOM. */
```
scwhittle: handle now?

upb/msg.c
```diff
@@ -258,83 +265,94 @@ static upb_msglayout *upb_msglayout_new(const upb_msgdef *m) {
...
     return NULL;
   }
 }
 
-upb_msgfactory *upb_msglayout_factory(const upb_msglayout *layout) {
-  return layout->factory;
+upb_msglayout *upb_msglayout_frominit_v1(
+    const struct upb_msglayout_msginit_v1 *init, upb_alloc *a) {
+  UPB_UNUSED(a);
+  /* If upb upgrades to a v2, this would create a heap-allocated v2. */
```
scwhittle: would it be good in debug mode to check that these paired methods are called correctly?
You could clone for example and then leak checker would catch missing uninit.

upb/msg.c
```diff
@@ -628,180 +644,150 @@ bool upb_visitor_visitmsg(upb_visitor *visitor, const upb_msg *msg) {
...
+  upb_inttable *extdict;
+  upb_msg_internal base;
+} upb_msg_internal_withext;
 
-    /* Use an 8-byte type to ensure all bytes are copied. */
-    if (!upb_inttable_init2(ext_dict, UPB_CTYPE_INT64, a)) {
-      upb_free(a, ext_dict);
-      return NULL;
-    }
+#define INTERNAL_MEMBERS_SIZE(l) \
```
scwhittle: this doesn't seem like it needs to be a macro
does the compiler not optimize this well?

upb/msg.c
```diff
@@ -150,63 +158,45 @@ static upb_msgval upb_msgval_fromdefault(const upb_fielddef *f) {
...
+  return l->data.fields[upb_fielddef_index(f)].hasbit;
 }
 
-static bool upb_msglayout_initdefault(upb_msglayout *l) {
-  const upb_msgdef *m = l->msgdef;
+static bool upb_msglayout_initdefault(upb_msglayout *l, const upb_msgdef *m) {
   upb_msg_field_iter it;
 
-  if (upb_msgdef_syntax(m) == UPB_SYNTAX_PROTO2 && l->size) {
+  if (upb_msgdef_syntax(m) == UPB_SYNTAX_PROTO2 && l->data.size) {
```
scwhittle: else return false?

upb/msg.c
```diff
@@ -219,7 +209,10 @@ static bool upb_msglayout_initdefault(upb_msglayout *l) {
       if (!upb_fielddef_isstring(f) &&
```
scwhittle: TODO for strings? or are they handled elsewhere?

upb/msg.c
```diff
@@ -628,180 +644,150 @@ bool upb_visitor_visitmsg(upb_visitor *visitor, const upb_msg *msg) {
...
+upb_msg *upb_msg_init(void *mem, const upb_msglayout *l, upb_alloc *a) {
+  upb_msg *msg = VOIDPTR_AT(mem, INTERNAL_MEMBERS_SIZE(l));
+  if (l->data.default_msg) {
+    memcpy(msg, l->data.default_msg, l->data.size);
   } else {
-    memset(msg, 0, l->size);
+    memset(msg, 0, l->data.size);
+  }
+
+  upb_msg_getinternal(msg)->alloc = a;
```
scwhittle: assert there was not an alloc (from default_msg)

upb/msg.c
haberman: For proto3 or a proto2 message with no fields, I think we can do nothing and return successfully.

upb/msg.c
haberman: Added TODO for strings.

upb/msg.c

upb/msg.c
haberman: These are unused (for now), so I removed them.

upb/msg.c

upb/msg.c

haberman:
Responded to comments, PTAL.

PR #88, opened 2017-07-05
First version of a real C codegen for upb.

haberman:
Also includes an implementation of the conformance tests
to display what the API usage will be like.

There is still a lot to do, and things that are broken (oneofs,
repeated fields, etc), but it's a good start.


tools/upbc.lua
```diff
@@ -1,20 +1,29 @@
...
 local upb = require "upb"
 
-local src = arg[1]
+local generate_upbdefs = false
+
+for _, argument in ipairs(arg) do
+  if argument == "--generate-upbdefs" then
+    generate_upbdefs = true
+  else
+    src = argument
```
scwhittle: assert that src isn't already set

upb/msg.h
```diff
@@ -405,21 +403,22 @@ typedef struct {
...
-  const struct upb_msglayout_msginit_v1 **submsgs;
+  const struct upb_msglayout_msginit_v1 *const* submsgs;
   const upb_msglayout_fieldinit_v1 *fields;
   const upb_msglayout_oneofinit_v1 *oneofs;
-  uint32_t *case_offsets;
   void *default_msg;
-  /* Must be aligned to 8.  Doesn't include internal members like unknown
-   * fields, extension dict, pointer to msglayout, etc. */
+  /* Must be aligned to sizeof(void*).  Doesn't include internal members like
+   * unknown * fields, extension dict, pointer to msglayout, etc. */
```
scwhittle: rm second *

upb/msg.c
```diff
@@ -343,7 +341,7 @@ static upb_msglayout *upb_msglayout_new(const upb_msgdef *m) {
 
   /* Size of the entire structure should be a multiple of its greatest
    * alignment. */
-  l->data.size = align_up(l->data.size, l->data.align);
+  l->data.size = align_up(l->data.size, 8 /* TODO: track for real? */);
```
scwhittle: merge todo with comment

upb/msg.h
```diff
@@ -110,6 +110,22 @@ const upb_visitorplan *upb_msgfactory_getvisitorplan(upb_msgfactory *f,
                                                      const upb_handlers *h);
 
 
+/** upb_stringview ************************************************************/
```
scwhittle: Seems like the stringview stuff could be split out into a separate commit

tests/conformance_upb.c
```diff
@@ -0,0 +1,162 @@
+/* This is a upb implementation of the upb conformance tests, see:
```
scwhittle: should this be part of the makefile?

tools/make_c_api.lua
```diff
@@ -0,0 +1,432 @@
...
+  [upb.TYPE_UINT64]   = "uint64_t",
+  [upb.TYPE_STRING]   = "upb_stringview",
+  [upb.TYPE_BYTES]    = "upb_stringview",
+}
+
+function strip_proto(filename)
+  return string.gsub(filename, '%.proto$','')
+end
+
+--[[
```
scwhittle: rm

tools/make_c_api.lua
```diff
@@ -0,0 +1,432 @@
...
+  append(' * file:\n')
+  append(' *\n')
+  append(' *     %s\n', filedef:name())
+  append(' *\n')
+  append(' * Do not edit -- your changes will be discarded when the file is\n')
+  append(' * regenerated. */\n\n')
+end
+
+local function field_layout_rank(field)
+  -- Order:
```
scwhittle: Is there a reason this is preferable to just using field num?

tools/make_c_api.lua
```diff
@@ -0,0 +1,432 @@
...
+    append('const upb_msglayout_msginit_v1 %s_msginit = {\n', msgname)
+    append('  %s,\n', submsgs_array_ref)
+    append('  %s,\n', fields_array_ref)
+    append('  NULL, /* TODO. oneofs */\n')
+    append('  NULL, /* TODO. default_msg */\n')
+    append('  UPB_ALIGNED_SIZEOF(%s), %s, %s, %s, %s\n',
+           msgname, field_count,
+           0, -- TODO: oneof_count
+           'false', -- TODO: extendable
+           'true' -- TODO: is_proto2
```
scwhittle: I think it would be better to explicitly fail compilation for now if these are set instead of generating something incomplete

upb/msg.h
haberman: Done, sent you a separate PR for that.

upb/msg.h
haberman: Done.

upb/msg.c

tools/make_c_api.lua
haberman: Done.

tools/make_c_api.lua
haberman: Added a comment with rationale.

tools/make_c_api.lua
haberman: The main problem for the TODO is that Lua doesn't currently have access to this information. A follow-up CL that is in progress fixes this.

tools/upbc.lua
haberman: Done.

tests/conformance_upb.c
haberman: As of this PR it doesn't work at all, so I didn't add it to the Makefile until the next (in-progress) CL.

tests/conformance_upb.c
```diff
@@ -0,0 +1,162 @@
...
+    case conformance_UNSPECIFIED:
+      fprintf(stderr, "conformance_upb: Unspecified output format.");
+      exit(1);
+
+    case conformance_PROTOBUF: {
+      size_t serialized_len;
+      char *serialized = protobuf_test_messages_proto3_TestAllTypes_serialize(
+          test_message, env, &serialized_len);
+      if (!serialized) {
+        fprintf(stderr, "conformance_upb: Error serialiing.");
```
haberman: Done.

haberman:
Responded to comments and resolved merge conflicts, PTAL.

upb/msg.c
```diff
@@ -342,8 +340,8 @@ static upb_msglayout *upb_msglayout_new(const upb_msgdef *m) {
   }
 
   /* Size of the entire structure should be a multiple of its greatest
-   * alignment. */
-  l->data.size = align_up(l->data.size, l->data.align);
+   * alignment.  TODO: track overall alignment for real? */
+  l->data.size = align_up(l->data.size, 8);
```
scwhittle: use sizeof(void*) instead of 8? that's what you have in the comment below.

haberman: Ah, that is something slightly different. The runtime requires that a msginit has its size aligned to `sizeof(void*)` because it will add internal members that are `void*`.  However the message itself might require even greater alignment if it has int64/uint64 members.

PR #90, opened 2017-07-17
Added new encoder/decoder that operates on upb_msglayout.

haberman:
This encoder/decoder still has some bugs, but 89% of the conformance tests pass, so it seems like a good start.


upb/encode.c
```diff
@@ -0,0 +1,381 @@
...
+
+#include "upb/upb.h"
+#include "upb/encode.h"
+#include "upb/structs.int.h"
+
+#define UPB_PB_VARINT_MAX_LEN 10
+#define CHK(x) do { if (!(x)) { return false; } } while(0)
+
+/* Maps descriptor type -> upb field type.  */
+static const uint8_t upb_desctype_to_fieldtype[] = {
```
scwhittle: factor this to be shared? It's copied a couple spots

upb/encode.c
```diff
@@ -0,0 +1,381 @@
...
+  UPB_TYPE_ENUM,            /* ENUM */
+  UPB_TYPE_INT32,           /* SFIXED32 */
+  UPB_TYPE_INT64,           /* SFIXED64 */
+  UPB_TYPE_INT32,           /* SINT32 */
+  UPB_TYPE_INT64,           /* SINT64 */
+};
+
+static size_t upb_encode_varint(uint64_t val, char *buf) {
+  size_t i;
+  if (val == 0) { buf[0] = 0; return 1; }
```
scwhittle: what about?
if (val <= 128) { buf[0] = val; return 1; }

upb/encode.c
```diff
@@ -0,0 +1,381 @@
...
+  return ret;
+}
+
+static bool upb_encode_growbuffer(upb_encstate *e, size_t bytes) {
+  size_t old_size = e->limit - e->buf;
+  size_t new_size = upb_roundup_pow2(bytes + (e->limit - e->ptr));
+  char *new_buf = upb_env_realloc(e->env, e->buf, old_size, new_size);
+  CHK(new_buf);
+
+  /* We want previous data at the end, realloc() put it at the beginning. */
```
scwhittle: is it better to use malloc then and potentially avoid an unnecessary copy?

upb/upb.h
```diff
@@ -623,7 +636,7 @@ void upb_env_uninit(upb_env *e);
 
 void upb_env_initonly(upb_env *e);
 
-upb_arena *upb_env_arena(upb_env *e);
+UPB_INLINE upb_arena *upb_env_arena(upb_env *e) { return (upb_arena*)e; }
```
scwhittle: I don't see removal of previous definition.

tests/conformance_upb.c
```diff
@@ -111,7 +114,7 @@ bool DoTestIo() {
   char *serialized_input;
   char *serialized_output;
   uint32_t input_size;
-  size_t output_size;
+  size_t output_size = 0;
```
haberman: Done.

tools/make_c_api.lua
```diff
@@ -129,25 +146,24 @@ end
...
   -- Order:
   --   1, 2, 3. primitive fields (8, 4, 1 byte)
-  --   4. oneof fields
-  --   5. string fields
-  --   6. submessage fields
-  --   7. repeated fields
+  --   4. string fields
+  --   5. submessage fields
+  --   6. repeated fields
   --
```
haberman: Done.

tools/make_c_api.lua
```diff
@@ -310,15 +346,46 @@ local function write_c_file(filedef, hfilename, append)
         submsg_set[field:subdef()] = true
       end
 
-      if has_hasbit(field) then
-        hasbit_indexes[field] = hasbit_count
-        hasbit_count = hasbit_count + 1
+      if field:containing_oneof() then
+        -- Do nothing now
```
haberman: Done.

tools/make_c_api.lua
```diff
@@ -310,15 +346,46 @@ local function write_c_file(filedef, hfilename, append)
...
+          hasbit_indexes[field] = hasbit_count
+          hasbit_count = hasbit_count + 1
+        end
+
+        append('  %s %s;\n', ctype(field), field:name())
       end
+    end
 
-      append('  %s %s;\n', ctype(field), field:name())
+    local oneof_last_fields = {}
```
haberman: Done.

tools/make_c_api.lua
```diff
@@ -350,17 +417,20 @@ local function write_c_file(filedef, hfilename, append)
...
-        local submsg_index = "-1"
+        local submsg_index = "UPB_NO_SUBMSG"
         local oneof_index = "UPB_NOT_IN_ONEOF"
         if field:type() == upb.TYPE_MESSAGE then
           submsg_index = submsg_indexes[field:subdef()]
         end
+        if field:containing_oneof() then
+          oneof_index = oneof_indexes[field:containing_oneof()]
+        end
         -- TODO(haberman): oneofs.
```
haberman: Done.

upb/decode.c
```diff
@@ -0,0 +1,572 @@
...
+                               const upb_msglayout_msginit_v1 *l);
+
+static bool upb_decode_varint(const char **ptr, const char *limit,
+                              uint64_t *val) {
+  uint8_t byte = 0x80;
+  int bitpos = 0;
+  const char *p = *ptr;
+  *val = 0;
+
+  while (byte & 0x80) {
```
haberman: Done.

upb/decode.c
```diff
@@ -0,0 +1,572 @@
...
+  upb_decode_setpresent(frame, field);
+  return true;
+}
+
+static bool upb_decode_fixedpacked(upb_array *arr, upb_stringview data,
+                                   int elem_size) {
+  int elements = data.size / elem_size;
+  void *field_mem;
+
+  CHK((data.size % elem_size) == 0);
```
haberman: Done.

upb/encode.c
haberman: Nice! Done.

upb/upb.h
haberman: I think I had failed to add a definition previously.

upb/encode.c
haberman: `realloc()` might be more memory efficient, which is a higher priority to me in this encoder/decoder. I'm not so worried these days about unnecessary copies, memcpy() is really fast compared to protobuf encoding/decoding.

upb/encode.c
haberman: I can't figure out what header file it should go into. I think I'll punt on this until there are more things of this sort, and it's more obvious where to put them.

upb/decode.c
```diff
@@ -0,0 +1,572 @@
...
+  new_bytes = new_size * arr->element_size;
+  new_data = upb_realloc(arr->alloc, arr->data, old_bytes, new_bytes);
+  CHK(new_data);
+
+  arr->data = new_data;
+  arr->size = new_size;
+  return true;
+}
+
+static void *upb_array_reserve(upb_array *arr, size_t elements) {
```
haberman: I would definitely do this if this function were made public. Since this is in the decoder only for now, it seems ok to take a more append-centric view and reserve the number of things you want to add.

haberman:
Responded to comments, PTAL.

upb/encode.c
```diff
@@ -258,7 +258,7 @@ static bool upb_encode_scalarfield(upb_encstate *e, const char *field_mem,
                                    bool is_proto3) {
 #define CASE(ctype, type, wire_type, encodeval) do { \
   ctype val = *(ctype*)field_mem; \
-  if (is_proto3 && val == 0) { \
+  if (is_proto3 && f->oneof_index == UPB_NOT_IN_ONEOF && val == 0) { \
```
scwhittle: could factor out
const bool omit_default_values = is_proto3 && f->oneof_index == UPB_NOT_IN_ONEOF;
and change all the ifs to use it.
Maybe there is a better variable name but I think it makes it easier to read

PR #92, opened 2017-09-19
Reserve unknown fields in upb

TeBoring:
1. For decoding, an unknownfields will be lazily created on message,
which contains bytes of unknown fields.
2. For encoding, if the unknownfields is present on message, all bytes
contained in it will be serialized.


upb/handlers.h
```diff
@@ -633,6 +634,10 @@ UPB_BEGIN_EXTERN_C
 /* Native C API. */
 
 /* Handler function typedefs. */
+typedef bool upb_addunknown_handlerfunc(void *c, const void *hd,
+                                        const char *buf, size_t n);
+typedef bool upb_unknown_handlerfunc(void *c, const void *hd, const char *buf,
```
haberman: What is the difference between these two? They are the same typedef, do they have any difference in meaning?

upb/pb/decoder.c
```diff
@@ -596,6 +599,14 @@ int32_t upb_pbdecoder_skipunknown(upb_pbdecoder *d, int32_t fieldnum,
     }
 
     if (d->top->groupnum >= 0) {
+      addunknown = (upb_addunknown_handlerfunc *)upb_handlers_gethandler(
```
haberman: Can't you replace all this code with upb_sink_putunknown()?

TeBoring: Done

upb/handlers.h
TeBoring: Removed

upb/pb/encoder.h
```diff
@@ -64,6 +64,8 @@ const upb_handlers *upb_pb_encoder_newhandlers(const upb_msgdef *m,
 upb_sink *upb_pb_encoder_input(upb_pb_encoder *p);
 upb_pb_encoder* upb_pb_encoder_create(upb_env* e, const upb_handlers* h,
                                       upb_bytessink* output);
+void upb_pb_encoder_encode_unknown(upb_pb_encoder* p, const char* buf,
```
haberman: I think this declaration and definition are unused?

TeBoring: Renamed to encode_unknown. This needs to be exposed for php extension to write custom handler to encode bytes into unknown fields.

upb/pb/encoder.c
```diff
@@ -374,6 +374,11 @@ static void *encode_startdelimfield(void *c, const void *hd) {
   return ok ? c : UPB_BREAK;
 }
 
+bool encode_unknown(void *c, const void *hd, const char *buf, size_t len) {
```
haberman: I think this should be static and not declared in .h?

TeBoring: This needs to be exposed to php extension to implement custom handler.

haberman:
This looks great. I'll merge as soon as the tests pass.

upb/pb/decoder.c
```diff
@@ -596,6 +596,7 @@ int32_t upb_pbdecoder_skipunknown(upb_pbdecoder *d, int32_t fieldnum,
     }
 
     if (d->top->groupnum >= 0) {
+      upb_sink_putunknown(&d->top->sink, d->checkpoint, d->ptr - d->checkpoint);
```
haberman: Please remove the TODO above.

Also I think this code is incorrect for the case of unknown groups. But it might be a little complicated to fix right now. Maybe just add a TODO that the code needs to be fixed for unknown groups.

haberman:
Besides the minor comment issue above, I think this is ready to merge. Thanks for your work on this!

upb/pb/decoder.c
TeBoring: done

PR #96, opened 2018-03-07
For encoding upb needs descriptor type instead of type.

TeBoring:


haberman:
Looks like the build failed?

TeBoring:
Fixed

PR #100, opened 2018-03-07
Fix oneof encoding/decoding

TeBoring:


upb/encode.c
```diff
@@ -360,7 +360,13 @@ bool upb_encode_message(upb_encstate* e, const char *msg,
       CHK(upb_encode_array(e, msg + f->offset, m, f));
     } else {
       if (upb_encode_hasscalarfield(msg, m, f)) {
-        CHK(upb_encode_scalarfield(e, msg + f->offset, m, f, !m->is_proto2));
+        if (f->oneof_index != UPB_NOT_IN_ONEOF) {
```
scwhittle: swap condition branches to remove negation

TeBoring: Fixed

PR #106, opened 2018-03-09
Prepare upb_value for encoding/decoding map.

TeBoring:


TeBoring:
jit tests still fail. Could you help?

haberman:
I think this is obsoleted by https://github.com/protocolbuffers/upb/pull/224.

PR #107, opened 2018-03-12
Update README.md

lauraguevara97:
fixed spelling mistake :)


haberman:
Thanks for the fix! I cringe at how out-of-date this README is in general. This PR is a good reminder that people are actually reading this sometimes, and I should update it...

PR #108, opened 2018-03-12
Store pointer to upb_sringview in upb_value, because uint64_t is smaller than upb_stringview

TeBoring:


haberman:
I think this is obsoleted by https://github.com/protocolbuffers/upb/pull/224.

PR #109, opened 2018-05-15
Allows the json parser to ignore unknown fields during parse

ebenoist:
This change makes it possible to add new fields to an message without
causing breaking changes in older clients.

Unblocks https://github.com/google/protobuf/pull/4304 and fixes https://github.com/google/protobuf/issues/4251

### DIFF
--- a/upb/json/parser.rl
+++ b/upb/json/parser.rl
@@ -107,6 +107,9 @@ struct upb_json_parsermethod {
 
   /* Keys are upb_msgdef*, values are upb_strtable (json_name -> fielddef) */
   upb_inttable name_tables;
+
+  /* Configurable value to drop unknown fields instead of returning a parse error */
+  bool ignore_unknown_fields;
 };
 
 #define PARSER_CHECK_RETURN(x) if (!(x)) return false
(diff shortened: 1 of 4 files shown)


TeBoring:
ping @haberman

upb/json/parser.rl
```diff
@@ -1036,11 +1039,13 @@ static bool end_membername(upb_json_parser *p) {
 
       return true;
     } else {
-      /* TODO(haberman): Ignore unknown fields if requested/configured to do
-       * so. */
-      upb_status_seterrf(&p->status, "No such field: %.*s\n", (int)len, buf);
-      upb_env_reporterror(p->env, &p->status);
-      return false;
+      if (p->method->ignore_unknown_fields) {
+        return false;
```
TeBoring: I think you forget to call ```multipart_end```

TeBoring: Have you tested this can ignore unknown?

ebenoist: This was tested as part of the original PR via the ruby interface. Is there a test suite here I can leverage?

https://github.com/google/protobuf/pull/4304/files

What does `multipart_end` do and why do I need to call it here?

TeBoring:
This should be no longer needed.

TeBoring:
Could you resolve the conflict?

PR #111, opened 2018-06-22
Adds ability to add extension ranges for a MessageDef.

hrsht:
This is the first of commits to support proto2 extensions.

Please let me know if I am on the right track with this.

cc: @yurokle

### DIFF
--- a/tests/test_def.c
+++ b/tests/test_def.c
@@ -431,6 +431,35 @@ static void test_oneofs() {
   upb_oneofdef_unref(o, &o);
 }
 
+static void test_extension_ranges() {
+  upb_status s = UPB_STATUS_INIT;
+  bool ok = true;
+  upb_symtab *symtab = upb_symtab_new(&symtab);
+  upb_msgdef *m = upb_msgdef_newnamed("TestMessage", &symtab);
+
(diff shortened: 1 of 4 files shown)


hrsht:
I signed it!

hrsht:
Ping on the review!

hrsht:
Closing this PR as it has been dormant for a while now.

haberman:
I apologize for my lack of response on this.

I've wanted upb to support extensions for a while, but there were always a lot of unanswered questions about how to do it. Also upb has been undergoing some pretty major changes in the last couple years. This is starting to settle down (especially once https://github.com/protocolbuffers/upb/pull/363 lands).

There are still a fair number of questions to answer, but I think your PR (giving `upb_msgdef` an array of extension ranges) is a good start.

`upb_msgdef` is immutable now, so this would only involve getters (not setters). I think we could pattern it somewhat after the C++ API: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.h#L381-L410

So feel free to re-open (or open a new PR). Otherwise hopefully I'll get a chance to do this myself sometime in the next few quarters.

PR #113, opened 2018-07-30
Support json parsing for wrapper values

TeBoring:


upb/json/parser.rl
```diff
@@ -1345,7 +1574,10 @@ static void end_object(upb_json_parser *p) {
     value
     <: any >{ fhold; fret; } ;
 
-  main := ws object ws;
+  main :=
+    ws
+    (object | number_object | true_object | false_object | string_object)
```
haberman: Why do you need to change "main"?  When do we need numbers or true/false/string at the top level?

To me number_object/true_object/false_object/etc. seem redundant since these cases are already handled in "value".

TeBoring: For top level wrapper values, there is no  wrapping ```{}```. Currently, the definition of ```object``` starts with ```{``` and ends with ```}```.  So it cannot  accept top level wrapper values.  I also have test in protobuf to verify this.

TeBoring:
ping?

upb/json/parser.rl
```diff
@@ -1325,27 +1457,103 @@ static void end_object(upb_json_parser *p) {
 
   value =
     number
-      >{ start_number(parser, p); }
-      %{ CHECK_RETURN_TOP(end_number(parser, p)); }
+      >{
+         if (is_top_level(parser)) {
```
haberman: Could we move all of this logic to start_number() instead? I want to keep the Ragel machine part of the file small and simple.

Likewise for the other values below.

upb/json/parser.rl
```diff
@@ -1325,27 +1457,103 @@ static void end_object(upb_json_parser *p) {
...
+           end_wrapper_object(parser);
+           if (!is_top_level(parser)) {
+             end_subobject(parser);
+           }
+         }
+       }
     | "true"
-      %{ CHECK_RETURN_TOP(parser_putbool(parser, true)); }
+      >{
+         if (is_top_level(parser)) {
```
haberman: Can we wait until the end of "true" to do all this work? It seems like all of the start/end code for boolean could go in parser_putbool().

Likewise for "false" below. We only care that we say true/false. We don't need to do anything at the beginning of the word.

PR #114, opened 2018-08-13
Enforced that upb_msg lives in an Arena only, and other simplifying.

haberman:
upb_msg was trying to be general enough that it could either live in
an arena or be allocated with malloc()/free().  This was too much
complexity for too little benefit.  We should commit to just saying
that upb_msg is arena-only.

I also ripped out the code to glue upb_msg to the existing
handlers-based encoder/decoder.  upb_msg has its own, small, simple
encoder/decoder.  I'm trying to whittle down upb_msg to a small
and simple core.

I updated the Lua extension for these changes.  Lua needs some more
work to properly create arenas per message.  For now I just created
a single global arena.

### DIFF
--- a/tests/bindings/lua/test_upb.pb.lua
+++ b/tests/bindings/lua/test_upb.pb.lua
@@ -27,25 +27,13 @@ local symtab = upb.SymbolTable{
 local factory = upb.MessageFactory(symtab);
 local TestMessage = factory:get_message_class("TestMessage")
 
-function test_decodermethod()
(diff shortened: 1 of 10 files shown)


haberman:
Travis failure is unrelated.

TeBoring:
Is arena recommend for all situation in cpp protobuf? My impression was in some situation using arena will cause peak memory usage. Does the problem also apply here?

TeBoring:
Cpp protobuf always do deep copy.  However, upb is for general usage. Some implementation may choose pass by reference.  Will that be a concern?

PR #115, opened 2018-08-17
Fix json parsing for timestamp

TeBoring:


upb/json/parser.rl
```diff
@@ -947,6 +973,134 @@ static bool start_stringval(upb_json_parser *p) {
...
+      }
+
+      nanos = val * 1000000000;
+      upb_gfree(nanos_buf);
+    } else {
+      i = 19;
+    }
+
+    /* Parse timezone */
+    if ((i == len - 1 && buf[i] == 'Z')) {
```
haberman: This is a lot of manual parsing. We should be using Ragel for this if we can.

Is it possible that our "value" machine can enter a special "timestamp" machine if we decide this is a timestamp field? Then we could specify the format of a timestamp in Ragel, and have functions that get called for the different parts of a timestamp that set the appropriate values.

upb/json/parser.rl
```diff
@@ -882,15 +904,23 @@ static bool end_bool(upb_json_parser *p, bool val) {
 
 static bool start_stringval(upb_json_parser *p) {
   if (is_top_level(p)) {
-    if (!is_string_wrapper_object(p)) {
+    if (is_string_wrapper_object(p)) {
+      start_wrapper_object(p);
+    } else if (is_timestamp_object(p)) {
+      start_object(p);
```
haberman: Can we have a separate start_timestamp() function that handles all of these timestamp cases? If timestamp is a separate machine it shouldn't call start_stringval() should it?

upb/json/parser.rl
```diff
@@ -947,6 +977,11 @@ static bool start_stringval(upb_json_parser *p) {
 static bool end_stringval_nontop(upb_json_parser *p) {
   bool ok = true;
 
+  if (is_timestamp_object(p)) {
```
haberman: Ditto, separate end_timestamp() function?

upb/json/parser.rl
TeBoring: start_stringval prepare the parser for parsing string data. We did the same for string wrapper. I think it makes sense to me because all string, string wrapper and timestamp have string json value. Doing this way also makes the ragel machine easier, otherwise we have to check stuff like whether it's top level message, whether it's timestamp in the ragel machine (basically copy the logic in start_stringval to ragel machine). Do you want me to change this?

PR #119, opened 2018-08-27
Bugfix for bigendian platforms by casting size_t to unint32_t

rohitsakala:
The reason for typecasting size_t to unint32_t is that size_t is of 8 bytes and uint32_t is only 4 bytes.
If not typecasted Memcpy fails to copy the *correct* four bytes on big-endian platforms.


rohitsakala:
I signed it!

TeBoring:
Does this fix https://github.com/protocolbuffers/protobuf/issues/4893?

rohitsakala:
Hi,

Yes. Haberman asked me to submit the PR here.

haberman:
Thanks for the fix!

PR #121, opened 2018-09-03
Changed C API to only define structs, a table, and a few minimal inline function.

haberman:
This makes C client code read more nicely. See conformance_upb.c for an example.

The downside is that it prevents us from changing the internal representation except to reorder struct fields.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,7 +148,9 @@ if(LUA_FOUND AND PROTOBUF_FOUND)
 
   add_custom_target(
     upbc ALL
-    DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/tools/upbc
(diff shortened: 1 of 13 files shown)


TeBoring:
I copied the amalgamated files to my php repository. But it's complaining: fatal error: upb/port_undef.inc: No such file or directory

TeBoring:
Can upb/port_def.inc and upb/port_undef.inc be treated like other header files?

TeBoring:
Or copy contents of upb/port_def.inc and upb/port_undef.inc to upb.h and upb.c?

haberman:
@TeBoring sorry for the breakage. Try https://github.com/google/upb/pull/125.

TeBoring:
Thanks for the quick fix!

haberman:
For sure. :) I'm assuming it worked? If so I'll merge it.

PR #126, opened 2018-09-10
Fix json encoding for wrappers, ListValue, Struct and Value.

TeBoring:


upb/def.h
```diff
@@ -794,6 +794,33 @@ class upb::MessageDef {
   /* Is this message a value? */
   bool value() const;
 
+  /* Is this message a doublevalue? */
```
haberman: Can we change this to use an enum for well-known types instead?  We shouldn't need separate accessors for every kind of well-known type.

enum upb_wellknowntype {
  UPB_WELLKNOWNTYPE_NONE = 0,  /* Not a well-known type. */
  UPB_WLELKNOWNTYPE_DOUBLEVALUE = 1,
  UPB_WELLKNOWNTYPE_FLOATVALUE = 2,
  /* etc. */
}

upb/json/parser.rl
```diff
@@ -1828,54 +1818,18 @@ static void end_object(upb_json_parser *p) {
...
-}
-
 static bool is_number_wrapper(const upb_msgdef *m) {
-  return is_double_value(m) ||
-         is_float_value(m) ||
-         is_int64_value(m) ||
-         is_uint64_value(m) ||
-         is_int32_value(m) ||
-         is_uint32_value(m);
+  return (upb_msgdef_wellknowntype(m) == UPB_WELLKNOWN_DOUBLEVALUE) ||
```
haberman: Could we put all of the wrapper types in a contiguous range so that this can just be something like (type >= UPB_WELLKNOWN_FIRSTWRAPPER && type <= UPB_WELLKNOWN_LASTWRAPPER)?

Also we should only call upb_msg_wellknowntype(m) once at a time, and store the value.

upb/json/parser.rl
```diff
@@ -2008,86 +1962,115 @@ static bool is_string_wrapper_object(upb_json_parser *p) {
...
-         is_bool_value(upb_fielddef_msgsubdef(p->top->f));
+         (upb_msgdef_wellknowntype(upb_fielddef_msgsubdef(p->top->f))
+              == UPB_WELLKNOWN_BOOLVALUE);
 }
 
 static bool does_boolean_wrapper_end(upb_json_parser *p) {
-  return p->top->m != NULL && is_bool_value(p->top->m);
+  return p->top->m != NULL &&
+         (upb_msgdef_wellknowntype(p->top->m)
+              == UPB_WELLKNOWN_BOOLVALUE);
```
haberman: A lot of functions are using this pattern. Create a function like this?

```
static bool top_is_wellknown(upb_json_parser *p, upb_wellknown_t type) {
  return p->top->m != NULL & upb_msgdef_wellknowntype(p->top->m) == type;
}
```

Then I don't think you would need all these separate functions.

upb/json/printer.c
```diff
@@ -906,16 +1070,49 @@ void printer_sethandlers(const void *closure, upb_handlers *h) {
     return;
   }
 
-  if (upb_msgdef_duration(md)) {
+  if (upb_msgdef_wellknowntype(md) == UPB_WELLKNOWN_DURATION) {
```
haberman: switch (upb_msgdef_wellknowntype(md)) ?

PR #127, opened 2018-09-12
Fix mac use local timezone for strptime by default.

TeBoring:


upb/json/parser.rl
```diff
@@ -1290,15 +1290,19 @@ static void start_timestamp_base(upb_json_parser *p, const char *ptr) {
...
   size_t len;
   const char *buf;
+  char timestamp_buf[23];
 
   if (!capture_end(p, ptr)) {
     return false;
   }
 
   buf = accumulate_getptr(p, &len);
+  memset(timestamp_buf, 0, 23);
```
haberman: please use sizeof(timestamp_buf) instead of hard-coded 23.

upb/json/parser.rl
```diff
@@ -1290,15 +1290,19 @@ static void start_timestamp_base(upb_json_parser *p, const char *ptr) {
...
+  memcpy(timestamp_buf, buf, len);
```
haberman: please check that len<sizeof(timestamp_buf).

upb/json/parser.rl
```diff
@@ -1290,15 +1290,19 @@ static void start_timestamp_base(upb_json_parser *p, const char *ptr) {
...
+  memcpy(timestamp_buf + 19, "GMT", 3);
```
haberman: Please use sizeof(timestamp_buf)-4, and mention NULL termination.

Also please explicitly NULL-terminate, instead of relying on the memset() above.

upb/json/parser.rl
TeBoring: Can I use a UPB_TIMESTAMP_BASE_SIZE (19) here?

upb/json/parser.rl
TeBoring: Can I use a UPB_TIMESTAMP_BASE_SIZE (19) here?

PR #130, opened 2018-10-30
Implement json decoding for Any message.

TeBoring:
type url may not appear as the first value in json. As a result,
other data cannot be resolved before resolving type url. To solve that,
this change caches the start and end position of unparsed values and
resolve them in end_any_object when type url has been resolved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,6 +191,9 @@ endif
 upb_json_SRCS = \
   upb/json/parser.c \
   upb/json/printer.c \
+  upb/pb/encoder.c \
+  upb/pb/varint.c \
+  upb/sink.c \
 
 # Ideally we could keep this uncommented, but Git apparently sometimes skews
 # timestamps slightly at "clone" time, which makes "Make" think that it needs
--- a/upb/json/parser.rl
+++ b/upb/json/parser.rl
@@ -34,6 +34,7 @@
 #include <time.h>
 
 #include "upb/json/parser.h"
+#include "upb/pb/encoder.h"
 
 #define UPB_JSON_MAX_DEPTH 64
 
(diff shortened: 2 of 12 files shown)


TeBoring:
Fix https://github.com/protocolbuffers/protobuf/issues/5130

upb/json/parser.rl
```diff
@@ -1060,6 +1246,50 @@ static bool start_stringval(upb_json_parser *p) {
...
+
+    packed = upb_symtab_lookupmsg2(p->symtab, buf, len);
+    if (packed == NULL) {
+      upb_status_seterrf(
+          &p->status, "Cannot find packed type: %.*s\n", (int)len, buf);
+      upb_env_reporterror(p->env, &p->status);
+      return false;
+    }
+
+    json_parser_any_frame_set_packed(p, p->top->any_frame, packed);
```
haberman: Can we use a name other than "packed"? I thought this was somehow related to packed repeated fields. Maybe `json_parser_any_frame_set_type()`? And maybe call the variable here `any_type`?

upb/json/parser.rl
```diff
@@ -75,15 +76,99 @@ static void end_structvalue_object(upb_json_parser *p);
...
+  const upb_handlers *encoder_handlers;
+  upb_pb_encoder *encoder;
+  upb_stringsink stringsink;
+
+  /* For decoding Any value field in json format. */
+  upb_json_parsermethod *parser_method;
+  upb_json_parser* parser;
+  upb_sink sink;
+
+  /* Mark the range of unpacked values in json input before type url. */
```
haberman: Maybe `uninterpreted` instead of `unpacked`?

Makefile
```diff
@@ -191,6 +191,9 @@ endif
 upb_json_SRCS = \
   upb/json/parser.c \
   upb/json/printer.c \
+  upb/pb/encoder.c \
```
haberman: We shouldn't put these sources in more than one library. Instead add dependencies on libupb.pb.a to binaries that need it.

Or help me get Kokoro Bazel tests going, so we can delete the Travis tests and the Makefile completely.

upb/json/parser.rl
TeBoring: done.

upb/json/parser.rl
TeBoring: changed to payload_type.

Makefile
TeBoring: Left this. Will fix it after kokoro fix.

PR #132, opened 2018-11-05
Added Bazel build, and auto-generate CMake build from it.

haberman:


TeBoring:
Do you have test to cover these build files?
Later, if a new file is added/deleted, how could it be found missing from those build files?

haberman:
> Do you have test to cover these build files?

I want to introduce Kokoro tests for this. That's why I added `kokoro/` files to this PR. The kokoro tests aren't enabled yet, but I want to do that in the next few days.

PR #141, opened 2018-12-10
Major refactoring/cleanup of upb

haberman:
This is a major cleanup of upb:

- `upb_refcounted` is gone. Nothing is refcounted anymore:
  - Now `upb_symtab` owns def objects, like how `proto2::DescriptorPool` owns `proto2::Descriptor` objects.
  - For handlers and decoder methods, there are top-level objects that own all of the handlers/methods for individual methods (`upb_handlercache`, `upb_pbcodecache`, `upb_jsoncodecache`).
- All defs are now build directly from generated descriptor.proto messages like `google_protobuf_FileDescriptor`.
  - Defs like `upb_msgdef` and `upb_fielddef` are now const objects only, and don't have mutators like `upb_fielddef_settype()`.
  - `upb_descreader` and `upb/glue.{c,h}` can now be deleted completely.
  - This change will also make it more viable to add complex descriptor features like custom options.
- There is no longer an inheritance hierarchy with `upb_def` being a supertype of `upb_msgdef` and `upb_enumdef`. The defs are now all distinct types.
- `upb_env` is gone, and `upb_arena` is simplified.
- The JIT is deleted. Nobody was actually using it. If we want to JIT in the future, it will write directly to generated object types, instead of a `upb_handlers`-based JIT.
- The C++ wrappers that used an unorthodox, arguably undefined-behavior way of wrapping has been changed to wrap in a normal way.
- Sink objects (`upb_sink`, `upb_bytessink`) are now by-value objects.  They are just two pointers, so there is no need to pass them by pointer.


tests/conformance_upb.c
```diff
@@ -39,7 +39,7 @@ void CheckedWrite(int fd, const void *buf, size_t len) {
   }
 }
 
-bool stringview_eql(upb_stringview view, const char *str) {
+bool strview_eql(upb_strview view, const char *str) {
```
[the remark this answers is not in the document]
haberman: Yes it's intended. It was one of the refactorings I put in this batch. The name was longer than necessary, but there is no deeper reason.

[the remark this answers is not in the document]
haberman: One of the goal of this refactoring was to change any APIs that I thought could be improved before gRPC starts using upb.

tools/make_cmakelists.py
```diff
@@ -59,27 +59,29 @@ def cc_binary(self, **kwargs):
...
-          add_custom_command(
-              TARGET %s POST_BUILD
-              COMMAND ${CMAKE_COMMAND} -E copy
-                      ${CMAKE_SOURCE_DIR}/%s
-                      ${CMAKE_CURRENT_BINARY_DIR}/%s)\n""" % (
-          kwargs["name"], data_dep, data_dep
-        ))
-
-    self._add_deps(kwargs)
+    # Disable this until we properly support upb_proto_library().
```
TeBoring: Is the commented out code intended?

upb/decode.h
```diff
@@ -7,10 +7,14 @@
 
 #include "upb/msg.h"
 
-UPB_BEGIN_EXTERN_C
```
TeBoring: Isn't this macro more convenient?

tools/make_cmakelists.py
haberman: Yes, for now. I didn't fix the CMake generator to replicate the somewhat complex logic of the Bazel upb_proto_library() rule. I would like to fix that at some point, but it's not worth it to do in this change.

upb/decode.h
haberman: The macro is a little bit nice, but clang-format doesn't know about it so it formats it incorrectly. Since the benefit it was providing is small, it didn't seem worth keeping.

I suppose clang-format might format it correctly if it looked more like a function, eg

```
UPB_BEGIN_EXTERN_C();
```

But overall I am trying to move away from things that are too clever.

PR #142, opened 2018-12-12
Add support for encoding Any in json

TeBoring:


upb/json/printer.c
```diff
@@ -920,6 +920,53 @@ static bool printer_endmsg_noframe(
...
+
+  const upb_fielddef* type_field = upb_msgdef_itof(md, UPB_ANY_TYPE);
+  const upb_fielddef* value_field = upb_msgdef_itof(md, UPB_ANY_VALUE);
+
+  upb_handlerattr empty_attr = UPB_HANDLERATTR_INITIALIZER;
+
+  /* type_url's json name is "@type" */
+  upb_handlerattr type_name_attr = UPB_HANDLERATTR_INITIALIZER;
+  upb_handlerattr value_name_attr = UPB_HANDLERATTR_INITIALIZER;
+  strpc *type_url_json_name = upb_gmalloc(sizeof(*type_url_json_name));
```
haberman: please create a helper function that creates a strpc from a null-terminated const char*. You can use this for both "@type" and "value".

upb/json/printer.c
```diff
@@ -920,6 +920,53 @@ static bool printer_endmsg_noframe(
...
+
+  /* Set up handlers. */
+  upb_handlers_setstartmsg(h, printer_startmsg, &empty_attr);
+  upb_handlers_setendmsg(h, printer_endmsg, &empty_attr);
+
+  upb_handlers_setstartstr(h, type_field, scalar_startstr, &type_name_attr);
+  upb_handlers_setstring(h, type_field, scalar_str, &empty_attr);
+  upb_handlers_setendstr(h, type_field, scalar_endstr, &empty_attr);
+
+  upb_handlers_setstartstr(h, value_field, scalar_startstr_onlykey,
```
haberman: Please add a comment that this is not the full and correct JSON encoding for Any, and that it requires further processing by the client based on the type URL.

TeBoring: done

upb/json/printer.c
TeBoring: done

PR #146, opened 2018-12-22
Fixed amalgamation to not list header files explicitly.

haberman:


haberman:
The macOS Kokoro failure is unrelated. It appears that the Mac Kokoro image doesn't have CMake available anymore.

PR #151, opened 2019-02-04
Adding support for big endian platform (s390x)

shahidhs-ibm:
These changes will fix following tests which currently are failing for big endian architecture (s390x):
1. test_varint
2. test_json

The changes made are:
1. tests/pb/test_varint.c => Correcting the input data for big endian. 
2. upb/pb/varint.int.h => Reversing the order of bytes to correct the data representation for big endian. 
3. upb/table.c => Choosing the MurmurHash2 over MurmurHashAligned2 as it is not working correctly fro s390x architecture at present.

### DIFF
--- a/upb/pb/varint.int.h
+++ b/upb/pb/varint.int.h
@@ -9,6 +9,7 @@
 #include <assert.h>
 #include <stdint.h>
 #include <string.h>
+#include <byteswap.h>
 #include "upb/upb.h"
 
 #ifdef __cplusplus
@@ -130,6 +131,9 @@ UPB_INLINE uint64_t upb_vencode32(uint32_t val) {
   uint64_t ret = 0;
   UPB_ASSERT(bytes <= 5);
   memcpy(&ret, buf, bytes);
+#ifdef UPB_BIG_ENDIAN
+  ret = bswap_64(ret);
+#endif
   UPB_ASSERT(ret <= 0xffffffffffU);
   return ret;
 }
--- a/upb/pb/varint.int.h
+++ b/upb/pb/varint.int.h
@@ -9,7 +9,6 @@
 #include <assert.h>
 #include <stdint.h>
 #include <string.h>
-#include <byteswap.h>
 #include "upb/upb.h"
 
 #ifdef __cplusplus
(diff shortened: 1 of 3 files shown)


shahidhs-ibm:
Any update?

haberman:
Hi there, it's great that you are getting upb working on s390x.

Unfortunately the test coverage for this repo is not very good right now. For example, here are some other places that will need to be fixed:

https://github.com/google/upb/blob/master/upb/pb/decoder.c#L468

https://github.com/google/upb/blob/master/upb/decode.c#L84

https://github.com/google/upb/blob/master/upb/encode.c#L95

I'm happy to accept the PR as-is (once it passes the Kokoro tests). Then we can work on further fixes if necessary for whatever you are doing.

upb/pb/varint.int.h
```diff
@@ -9,6 +9,7 @@
 #include <assert.h>
 #include <stdint.h>
 #include <string.h>
+#include <byteswap.h>
```
haberman: This isn't a portable header. Also I think modern compilers don't need to the special intrinsic -- they can recognize the pattern. Let use some code like this instead:

https://godbolt.org/z/Nfj6qT

shahidhs-ibm:
Thanks @haberman 
I am working on code changes that you have suggested as well as fixing failures of test_encoder/test_decoder tests that are failing for s390x. Will update you soon.

upb/pb/varint.int.h
shahidhs-ibm: This looks better?
```
# git diff ../upb/pb/varint.int.h
diff --git a/upb/pb/varint.int.h b/upb/pb/varint.int.h
index 5de1296..752a579 100644
--- a/upb/pb/varint.int.h
+++ b/upb/pb/varint.int.h
@@ -9,7 +9,6 @@
 #include <assert.h>
 #include <stdint.h>
 #include <string.h>
-#include <byteswap.h>
 #include "upb/upb.h"

 #ifdef __cplusplus
@@ -25,6 +24,18 @@ extern "C" {
  * descriptor type (upb_descriptortype_t). */
 extern const uint8_t upb_pb_native_wire_types[];

+UPB_INLINE uint64_t byteswap64(uint64_t val)
+{
+return ((((val) & 0xff00000000000000ull) >> 56)
+   | (((val) & 0x00ff000000000000ull) >> 40)
+   | (((val) & 0x0000ff0000000000ull) >> 24)
+   | (((val) & 0x000000ff00000000ull) >> 8)
+   | (((val) & 0x00000000ff000000ull) << 8)
+   | (((val) & 0x0000000000ff0000ull) << 24)
+   | (((val) & 0x000000000000ff00ull) << 40)
+   | (((val) & 0x00000000000000ffull) << 56));
+}
+
 /* Zig-zag encoding/decoding **************************************************/

 UPB_INLINE int32_t upb_zzdec_32(uint32_t n) {
@@ -132,7 +143,7 @@ UPB_INLINE uint64_t upb_vencode32(uint32_t val) {
   UPB_ASSERT(bytes <= 5);
   memcpy(&ret, buf, bytes);
 #ifdef UPB_BIG_ENDIAN
-  ret = bswap_64(ret);
+  ret = byteswap64(ret);
 #endif
   UPB_ASSERT(ret <= 0xffffffffffU);
   return ret;
```

shahidhs-ibm:
@haberman Could you please review the changes and check if they are passing Kokoro tests?

shahidhs-ibm:
@haberman Kokoro tests for Ubuntu are passing, however, I believe that they are failing for macOS for reason unrelated to this PR. Could you please check and confirm?

haberman:
Looks good, thanks for the contribution. :)

PR #152, opened 2019-03-05
Fix the issue for strptime

TeBoring:
* Define _XOPEN_SOURCE
* Exclude usage on windows


TeBoring:
Not sure why mac fails. @haberman could you help take a look?

haberman:
Looks like it's a change in the newest Bazel. I'll debug this separately. Merging for now.

PR #153, opened 2019-03-12
Fix parsing repeated well known type value

TeBoring:


upb/json/parser.rl
```diff
@@ -2002,6 +2012,7 @@ static bool start_subobject(upb_json_parser *p) {
     inner->m = upb_fielddef_msgsubdef(p->top->f);
     set_name_table(p, inner);
     inner->f = NULL;
+    inner->is_repeated = false;
```
haberman: Can you factor out the code that sets the inner frame to default values?  Something like:

inner = init_frame(p);

PR #155, opened 2019-03-20
Changed enums to be open int32_t.

haberman:
This also fixes a bug where generated code wouldn't
compile if the field's enum was defined in another file.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -237,10 +237,19 @@ cc_test(
 
 proto_library(
     name = "test_json_proto",
-    srcs = ["tests/json/test.proto"],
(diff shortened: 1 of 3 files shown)


BUILD
```diff
@@ -237,10 +237,19 @@ cc_test(
 
 proto_library(
     name = "test_json_proto",
-    srcs = ["tests/json/test.proto"],
+    srcs = [
+        "tests/json/test.proto",
+        "tests/json/enum_from_separate_file.proto",
```
TeBoring: Is this file added in this PR?

PR #173, opened 2019-05-23
Update upb to be forward compatible with Bazel 0.27

hlopko:
Specifically, this PR prepares upb for:

https://github.com/bazelbuild/bazel/issues/7793
https://github.com/bazelbuild/bazel/issues/7260


haberman:
I have no idea why the macOS Kokoro build ran on Bazel 0.21. I have to assume this is spurious.

PR #188, opened 2019-06-25
Create README for fuzzing corpus folder 

Google-Autofuzz:


BUILD
```diff
@@ -363,12 +362,13 @@ cc_test(
...
 cc_binary(
-    testonly = 1,
     name = "file_descriptor_parsenew_fuzzer",
+    testonly = 1,
     srcs = ["tests/file_descriptor_parsenew_fuzzer.cc"],
     copts = CPPOPTS + select({
         "//conditions:default": [],
-        ":fuzz": ["-fsanitizer=fuzzer,address"],
+        ":fuzz": [],
+        #       ":fuzz": ["-fsanitizer=fuzzer,address"],
```
haberman: I don't think you want this to be empty?

Google-Autofuzz: You are right that should not be empty. I am trying to fix the last error message. The only error I am getting now in Travis CI when PR to Google/OSS-Fuzz was during the libfuzzer undefined test. It looks like the fuzz target wasn't built in "--sanitized undefined" config. So far everything works for libfuzzer address and memory. 

https://travis-ci.org/google/oss-fuzz/jobs/550517792#L2272

tests/corpus/temp.cc
```diff
@@ -0,0 +1 @@
+// Hello World
```
haberman: What is the purpose of this .cc file? Can we remove it?

I assume the corpus directory is for storing the fuzz corpus? Can we rename it `tests/fuzz_corpus`?

Google-Autofuzz: Yes, you can remove it and rename it to `tests/fuzz_corpus`.

PR #193, opened 2019-07-17
Disabling gcc flags for Windows build.

nicolasnoble:


BUILD
```diff
@@ -68,7 +73,10 @@ cc_library(
         "upb/encode.h",
         "upb/upb.h",
     ],
-    copts = COPTS,
```
haberman: Can we put the select inside the `COPTS` var so we don't need to repeat it for every cc_library()?

bazel/upb_proto_library.bzl
```diff
@@ -175,7 +175,7 @@ def _upb_proto_rule_impl(ctx):
              "_WrappedGeneratedSrcsInfo (aspect should have handled this).")
     cc_info = dep[_WrappedCcInfo].cc_info
     srcs = dep[_WrappedGeneratedSrcsInfo].srcs
-    lib = cc_info.linking_context.libraries_to_link.to_list()[0]
+    lib = cc_info.linking_context.libraries_to_link[0]
```
haberman: `to_list()` should be working I think? It appears to exist way back to Bazel 0.20 and before. What breaks when you have `to_list()` in there?

BUILD
nicolasnoble: I don't think we can, no. I can try.

bazel/upb_proto_library.bzl
nicolasnoble: That doesn't work with bazel 0.26, no. It says that `linking_context.libraries_to_link` doesn't have it `to_list()`:

```
type 'list' has no method to_list()
```

BUILD
nicolasnoble: That's a no by the way. Looks like select can only work either as an actual property, or as the return value of a skylark macro. We *could* push it into a skylark macro, something along these lines:

```python
def get_copts():
  return select({
    "windows": [],
    "//condition:default": ["-Werror"],
  })

bazel/upb_proto_library.bzl
nicolasnoble: Managed to wrangle some code to make it work pre and post bazel 0.27.

BUILD
nicolasnoble: I'm going to go ahead and merge as is, since there's no direct way to do what you were suggesting. There's a reason why we wrap all of our libraries and binaries in macros on grpc :)

PR #202, opened 2019-08-08
Accpet string for numeric wrappers

TeBoring:


TeBoring:
The broken mac test seems not related

PR #216, opened 2019-10-22
Use memcpy to perform unaligned reads

XrXr:
Creating and reading from unaligned pointers is UB and I'm trying to
run upb on a platform (GraalVM) that is sensitive to that unfortunately.
Recent compilers are smart enough to fold the memcpy down to a simple
memory load on platforms that support it, so this should mostly be a
aesthetic change.


XrXr:
By the way, if you are interested in removing some code, this new version of murmur runs on platforms that don't allow unaligned reads so it should be able to replace the other, more complicated version that avoids unaligned read.

PR #219, opened 2019-10-30
Fixed compiler error "error: control reaches end of non-void function".

haberman:


g-easy:
Thanks! This fixes `bazel build --copt -fsanitize=address :upb` for me.

PR #224, opened 2019-11-12
Implemented support for maps (generated & reflection), and improved reflection API

haberman:
This is a large change, and contains several major improvements:

New public APIs (needs the most thorough review):
1. `upb/reflection.h` is now upb's public reflection interface.
1. Maps fields now have a map-based generated code API (previously their generated API was a repeated field of entry).
1. Message and field descriptors `upb_msgdef` and `upb_fielddef` now contain and expose a pointer to the tables that upb uses for parsing and serialization:
    - `const upb_msglayout *upb_msgdef_layout()` returns a message table
    - `const upb_msglayout_field *upb_fielddef_layout()` returns a field layout.
1. Note: the format of these parsing tables is still private. All a user can do is pass the table to another upb function like `upb_decode()` or `upb_encode()`.

There are a lot of implementation changes in this PR:
1. Map fields now now have a true map representation in messages.
1. Parsing and serialization can write/read maps respectively, and reflection can access them too.
1. The code for generating a layout dynamically at runtime was moved into `upb/def.c`. The separate `upb_msgfactory` type is no longer required.
1. When a descriptor for a generated message is loaded, it explicitly passes in a reference to the generated parsing tables. This means that generated reflection `foo.upbdefs.c` now depends on the generated tables in `foo.upb.c`, which required reworking the Bazel aspects a bit so one aspect can depend on another.
1. The Lua extension has been extensively rewritten to use the new reflection interface. The Lua protobuf library here is not particularly meant for end-users, it just provides a convenient way of validating the reflection interface and implementation.
1. The Lua extension is now a very capable protobuf implementation supporting messages, repeated fields, maps, oneofs, and reflection. Because I'm especially eager to show that Ruby/PHP could be rewritten on top of these APIs, it follows the Ruby/PHP model where submessage assignment is allowed.
1. The Lua tests are reworked and now build into a single static binary instead of putting the upb extension in a separate .so -- this makes test failures easier to reproduce and debug.


TeBoring:
Took a look of the new map implementation and reflection API. In general, it should work for php. Looking into more details.

upb/def.c
```diff
@@ -819,6 +829,193 @@ void upb_oneof_iter_setdone(upb_oneof_iter *iter) {
...
+  }
+  UPB_UNREACHABLE();
+}
+
+static uint8_t upb_msg_fielddefsize(const upb_fielddef *f) {
+  if (upb_msgdef_mapentry(upb_fielddef_containingtype(f))) {
+    /* Map entries aren't actually stored, they are only used during parsing.
+     * For parsing, it helps a lot if all map entry messages have the same
+     * layout. */
+    return sizeof(upb_strview);
```
TeBoring: Add a reference to where may_entry is defined?

upb/def.c
```diff
@@ -819,6 +829,193 @@ void upb_oneof_iter_setdone(upb_oneof_iter *iter) {
...
+    }
+
+    /* TODO: we probably should sort the fields by field number to match the
+     * output of upbc, and to improve search speed for the table parser. */
+    f->layout_index = f->index_;
+
+    if (upb_fielddef_issubmsg(f)) {
+      const upb_msgdef *subm = upb_fielddef_msgsubdef(f);
+      field->submsg_index = submsg_count++;
+      submsgs[field->submsg_index] = subm->layout;
```
TeBoring: Is this guaranteed to be initialized here?

haberman: Yes, see line 1508, with comment:

```
/* Allocate now (to allow cross-linking), populate later. */
```

upb/def.c
haberman: Done.

TeBoring: Sorry, what is updated here? I want reference here because I couldn't easily figure out why it is sizeof(upb_strview).
Maybe just create an alias "map_entry_field"?

haberman: Oops, sorry I forgot to push. PTAL.

upb/reflection.h
```diff
@@ -0,0 +1,126 @@
...
+
+/** upb_array *****************************************************************/
+
+/* Creates a new array on the given arena that holds elements of this type. */
+upb_array *upb_array_new(upb_arena *a, upb_fieldtype_t type);
+
+/* Returns the size of the array. */
+size_t upb_array_size(const upb_array *arr);
+
+/* Returns the given element, which must be within the array's current size. */
```
TeBoring: Do we also need one for mutable message?

PR #227, opened 2019-11-15
bazel: use canonical repository name for absl

lizan:
https://github.com/abseil/abseil-cpp/blob/master/WORKSPACE#L17


haberman:
Does the current code cause problems in practice?

lizan:
@haberman yes, when a project depends on both absl and upb.

PR #228, opened 2019-11-19
Mirror zlib's http_archive

philwo:
Context: https://github.com/bazelbuild/bazel/issues/10270


haberman:
If you re-sync I can merge this.

PR #242, opened 2020-02-01
Text format serializer for upb_msg

haberman:
This implements a text format serializer `upb_textencode()` for `upb_msg` messages. It encodes into a fixed-size buffer, and if the output is too big it truncates. The output is always NULL-terminated, and the return value indicates the number of bytes that would have been written. These are the same semantics as `snprintf()`.

### DIFF
--- a/upb/textencode.c
+++ b/upb/textencode.c
@@ -5,7 +5,7 @@
  * string every time, and it allocates memory for every put.
  */
 
-#include "upb/pb/textprinter.h"
+#include "upb/textencode.h"
 
 #include <ctype.h>
 #include <float.h>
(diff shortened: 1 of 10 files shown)


upb/textencode.c
```diff
@@ -0,0 +1,393 @@
...
+
+    txtenc_indent(e);
+    txtenc_printf(e, "%s: {", upb_fielddef_name(f));
+    txtenc_endfield(e);
+    e->indent_depth++;
+
+    txtenc_field(e, key, key_f);
+    txtenc_field(e, val, val_f);
+
+    e->indent_depth++;
```
gerben-s: --?

PR #261, opened 2020-03-31
Add vcpkg installation instructions

PhoebeHui:
upb is available as a port in VCPKG , documenting the install process here will help users get started by providing a single set of commands to build upb, ready to be included in their projects.

VCPKG is a C++ library manager that simplifies installation for upb and other project dependencies, we also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.


haberman:
Hi, thanks for taking the effort to package upb in vcpkg.

I want to be clear that upb doesn't currently provide ABI stability, and even the API is still changing some in backward-incompatible ways. For this reason, I've generally preferred not to have an "install" target for CMake.

Do vcpkg users expect that libraries are providing a stable API and ABI?

PhoebeHui:
@haberman, thanks for your feedback!

Here is the [patch](https://github.com/microsoft/vcpkg/blob/master/ports/upb/add-cmake-install.patch) that supports to find_package(upb) in vcpkg.

It would be convinient to use t the cmake targets in vcpkg, the revision for upb kept update by cntributor and team members, upb need rebuild and installed when the revision changed. So I think installing cmake targets would be fine.

We didn't get any feedback from users that reqiures a stable API and ABI by now.

PR #260, opened 2020-03-23
Update Protobuf to v3.11.4 and Abseil to LTS 2020-02-25

Yannic:

PR #264, opened 2020-04-05
Updated to new protobuf version, and added support for packed=false.

haberman:
This is a continuation of https://github.com/protocolbuffers/upb/pull/260.

I fixed the conformance test failures by respecting `packed = true` and `packed = false` (before we would unconditionally pack everything).

Also fixed one bug when a message map value is omitted from the wire.


Yannic:
@googlebot I consent.

PR #270, opened 2020-04-21
Support for proto3 optional.

haberman:
The new methods on descriptor are designed to mimic the ones in C++.

There are no tests for the generated code or reflection yet. So far this is just enough to expose the information about proto3 optional to Ruby.

### DIFF
--- a/generated_for_cmake/upb/json/parser.c
+++ b/generated_for_cmake/upb/json/parser.c
@@ -735,7 +735,7 @@ static bool multipart_text(upb_json_parser *p, const char *buf, size_t len,
 /* Note: this invalidates the accumulate buffer!  Call only after reading its
  * contents. */
 static void multipart_end(upb_json_parser *p) {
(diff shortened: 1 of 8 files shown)


generated_for_cmake/upb/json/parser.c
```diff
@@ -735,7 +735,7 @@ static bool multipart_text(upb_json_parser *p, const char *buf, size_t len,
 /* Note: this invalidates the accumulate buffer!  Call only after reading its
  * contents. */
 static void multipart_end(upb_json_parser *p) {
-  UPB_ASSERT(p->multipart_state != MULTIPART_INACTIVE);
+  /* UPB_ASSERT(p->multipart_state != MULTIPART_INACTIVE); */
```
TeBoring, line 738: Remove?

haberman: I'm leaving it because it was false sometimes, so it causes a run with assertions to fail. But I want to preserve it because it was theoretically supposed to be true. I'll add a comment to explain.

PR #272, opened 2020-04-26
Optimized / Simplified arena

haberman:
Benchmarks show improvement in arena create/destroy. Unfortunately there is an unexplainable regression in the parse benchmark, even when the arena is not creating any heap blocks.


haberman:
Reverting this as I split it apart and merged separately.

PR #302, opened 2020-07-06
Verify UTF-8 when parsing proto3 string fields.

haberman:


upb/decode.c
```diff
@@ -155,6 +157,40 @@ static const char *decode_msg(upb_decstate *d, const char *ptr, upb_msg *msg,
...
+    offset = utf8_offset[(uint8_t)buf[i]];
+    if (offset == 0 || i + offset > len) {
+      decode_err(d);
+    }
+    for (j = i + 1; j < i + offset; j++) {
+      if ((buf[j] & 0xc0) != 0x80) {
+        decode_err(d);
+      }
+    }
+    i += offset;
```
gerben-s, line 189: I think this code is a tad naive in it's performance. If that is an issue at the moment

haberman: Would love to revisit later.

PR #314, opened 2020-10-12
Fixed bug in addunknown and added ASAN poisoning.

haberman:
ASAN poisoning will help detect errors where users are reading/writing past their arena allocation.

For now we don't poison initial blocks, because then we would have to track them.

The next step would be to track the `upb_alloc*` of each block, so initial blocks can have a `upb_alloc*` of `NULL`. This will also solve the bug of fusing arenas with different allocators.

ASAN poisoning showed a bug in `_upb_msg_addunknown`, which I fixed.

### DIFF
--- a/upb/port_def.inc
+++ b/upb/port_def.inc
@@ -182,7 +182,6 @@ int msvc_vsnprintf(char* s, size_t n, const char* format, va_list arg);
 #define UPB_NAN (0.0 / 0.0)
 #endif
 
-// User code should use macros instead of functions.
 #if defined(__SANITIZE_ADDRESS__)
 #define UPB_ASAN 1
 #ifdef __cplusplus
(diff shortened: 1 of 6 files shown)


upb/port_def.inc
```diff
@@ -181,3 +181,26 @@ int msvc_vsnprintf(char* s, size_t n, const char* format, va_list arg);
 #else
 #define UPB_NAN (0.0 / 0.0)
 #endif
+
+// User code should use macros instead of functions.
```
gerben-s, line 185: what do you mean here? I would think user code should never use UPB macros.

upb/port_undef.inc
```diff
@@ -26,3 +26,5 @@
 #undef _upb_snprintf
 #undef _upb_vsnprintf
 #undef _upb_va_copy
+#undef UPB_POISON_MEMORY_REGION
+#undef UPB_UNPOISON_MEMORY_REGION
```
gerben-s, line 30: UPB_ASAN as well?

gerben-s (approved):

upb/port_undef.inc
haberman: Done.

upb/port_def.inc
haberman: Bad copy/paste, removed.

PR #323, opened 2020-10-17
Split monolithic BUILD file into many build files.

haberman:


gerben-s (approved):
I think its great to organize the BUILD better

PR #326, opened 2020-10-28
Drop C89/C90 support and MSVC prior to Visual Studio 2015.

haberman:
upb previously attempted to support C89 and pre-2015 versions
of Visual Studio. This was to support older compilers with
limited C99 support (particularly MSVC). But as of last August,
even gRPC has dropped support for MSVC prior to 2015
https://github.com/grpc/grpc.io/commit/c87276d058d4a8464204bd12b7892f0f3bf0071d

Therefore it seems safe for upb to no longer attempt C89 support
(we were already not truly C89 compliant, with our use of "bool").
We now explicitly require C99 or greater and MSVC 2015 or greater.

This cleaned up port_def.inc a fair bit. I took the chance to
also remove some obsolete macros.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -6,8 +6,8 @@ load(
 load(
     "//bazel:upb_proto_library.bzl",
     "upb_proto_library",
-    "upb_proto_reflection_library",
     "upb_proto_library_copts",
+    "upb_proto_reflection_library",
 )
 
 # copybara:strip_for_google3_begin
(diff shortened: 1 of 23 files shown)


nicolasnoble:
Sure, that's fair.

nicolasnoble (approved):

gerben-s (approved):

veblush (approved):

PR #342, opened 2020-11-15
Clamp 32-bit varints to 5 bytes to fix a fuzz failure.

haberman:


upb/decode.c
```diff
@@ -221,11 +221,18 @@ static const char *decode_varint64(upb_decstate *d, const char *ptr,
 UPB_FORCEINLINE
 static const char *decode_varint32(upb_decstate *d, const char *ptr,
                                    uint32_t *val) {
-  uint64_t u64;
-  ptr = decode_varint64(d, ptr, &u64);
-  if (u64 > UINT32_MAX) decode_err(d);
-  *val = (uint32_t)u64;
-  return ptr;
+  uint64_t byte = (uint8_t)*ptr;
```
gerben-s, line 224: I don't think that this according to the spec. 

What you need is a decode_tag that does not read more than 5 bytes

haberman: This function is only used for tags and delimited field lengths.

gerben-s: perhaps a different name to reflect that, but that's up to you. It's definitely an error to accidentally use these for parsing fields.

haberman: I changed it to `decode_tag()` and switched parsing of delimited lengths to use `decode_varint64()` instead.

It makes me realize the spec is not clear about how many bytes of over-encoded varint to accept in different contexts.

gerben-s (approved):

PR #363, opened 2021-01-11
Deleted the legacy "Handlers" APIs. upb can finally be deserving of its name.

haberman:
This is possible now that all users have been migrated to the new upb_msg APIs.


gerben-s (approved):
Congrats

PR #364, opened 2021-01-11
Adds CIFuzz

catenacyber:
Now that upb got integrated into oss-fuzz by https://github.com/google/oss-fuzz/pull/4914, it is possible to have CIFuzz cf https://google.github.io/oss-fuzz/getting-started/continuous-integration/

CIFuzz is a GitHub action/CI job that runs the project's fuzz targets on pull requests, and so can prevent bugs from reaching the main branch.

I'll be happy to provide more info if needed

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,29 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+ Fuzzing:
+   runs-on: ubuntu-latest
+   strategy:
+     fail-fast: false
+     matrix:
+       sanitizer: [address, undefined]
+   steps:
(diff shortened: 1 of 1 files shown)


catenacyber:
Friendly ping on this PR @haberman ?

haberman:
Thanks for this, sorry I missed it before.

Do I need to do anything else to activate the GitHub action? Or does merely having the file in the source tree activate the GitHub action?

catenacyber:
Thanks.

>Do I need to do anything else to activate the GitHub action? Or does merely having the file in the source tree activate the GitHub action?

Better to test to be sure.
It looks like CIFuzz is indeed running for #369 nothing more needs to be done
We will see if it catches a bug (the removed check looked promising but I am not sure)

PR #369, opened 2021-02-15
Dummy test for CIFuzz

catenacyber:
Not to be merged !

cf https://github.com/protocolbuffers/upb/pull/364#issuecomment-779398217


catenacyber:
Woohoo fuzzers found a way to crash !

PR #376, opened 2021-03-04
Fixed quadratic memory usage in upb_array_append().

haberman:
We were erroneously calling realloc() instead of resize(),
forcing the entire array to be reallocated for every array
append.


fowles (approved):
LGTM

PR #378, opened 2021-03-29
Allow arena fuse to fail

fowles:
Track initial blocks to avoid having fuse operate on arenas that cannot
be fused.

### DIFF
--- a/upb/upb.c
+++ b/upb/upb.c
@@ -67,6 +67,18 @@ static void *upb_global_allocfunc(upb_alloc *alloc, void *ptr, size_t oldsize,
   }
 }
 
+static uint32_t* upb_cleanup_pointer(void* cleanup_metadata) {
+  return (uint32_t*)((uintptr_t)cleanup_metadata & ~0x1);
+}
+
+static bool upb_cleanup_has_initial_arena(void* cleanup_metadata) {
+  return (uintptr_t)cleanup_metadata & 0x1;
+}
+
+static void* upb_cleanup_metadata(uint32_t* cleanup, bool has_initial_arena) {
+  return (void*)((uintptr_t)cleanup | has_initial_arena);
+}
+
 upb_alloc upb_alloc_global = {&upb_global_allocfunc};
 
 /* upb_arena ******************************************************************/
@@ -112,7 +124,8 @@ static void upb_arena_addblock(upb_arena *a, upb_arena *root, void *ptr,
 
   a->head.ptr = UPB_PTR_AT(block, memblock_reserve, char);
   a->head.end = UPB_PTR_AT(block, size, char);
-  a->cleanups = &block->cleanups;
+  a->cleanup_metadata = upb_cleanup_metadata(
+      &block->cleanups, upb_cleanup_has_initial_arena(a->cleanup_metadata));
 
   UPB_POISON_MEMORY_REGION(a->head.ptr, a->head.end - a->head.ptr);
 }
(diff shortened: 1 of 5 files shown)


upb/upb.int.h
```diff
@@ -9,7 +9,10 @@ typedef struct mem_block mem_block;
 
 struct upb_arena {
   _upb_arena_head head;
-  uint32_t *cleanups;
+  /* Stores cleanup metadata for this arena.
+   * - a pointer to the current cleanup counter.
+   * - a boolean indicating if there is an unowned initial block.  */
+  void *cleanup_metadata;
```
haberman, line 15: If we're pointer tagging let's use `uintptr_t`.

upb/upb.c
```diff
@@ -67,6 +67,18 @@ static void *upb_global_allocfunc(upb_alloc *alloc, void *ptr, size_t oldsize,
   }
 }
 
+static uint32_t* upb_cleanup_pointer(void* cleanup_metadata) {
+  return (uint32_t*)((uintptr_t)cleanup_metadata & ~0x1);
+}
+
+static bool upb_cleanup_has_initial_arena(void* cleanup_metadata) {
```
haberman, line 74: `has_initial_block` instead of `has_initial_arena`?

upb/upb.c
```diff
@@ -67,6 +67,18 @@ static void *upb_global_allocfunc(upb_alloc *alloc, void *ptr, size_t oldsize,
   }
 }
 
+static uint32_t* upb_cleanup_pointer(void* cleanup_metadata) {
```
haberman, line 70: upb puts pointers on the right, like LLVM. I should add a `.clang-format` for this.

haberman (changes requested):

upb/upb.c
```diff
@@ -239,11 +255,13 @@ bool upb_arena_addcleanup(upb_arena *a, void *ud, upb_cleanup_func *func) {
   return true;
 }
 
-void upb_arena_fuse(upb_arena *a1, upb_arena *a2) {
+bool upb_arena_fuse(upb_arena *a1, upb_arena *a2) {
   upb_arena *r1 = arena_findroot(a1);
   upb_arena *r2 = arena_findroot(a2);
 
-  if (r1 == r2) return;  /* Already fused. */
+  if (r1 == r2) return true;  /* Already fused. */
```
haberman, line 262: Maybe it would be simplest and safest right now if we also make the fuse fail if either arena has a value other than `&upb_alloc_global` for the allocator.

Any custom allocator could have the same problem as the initial block: it could have a hard lifetime bound that cannot be extended arbitrarily.

fowles: I thought about this, but anything that presents an allocator interface really is signing up to let things live until the free function is called.

upb/upb.c
fowles: Done

upb/upb.c
fowles: Done

upb/upb.int.h
fowles: Done

upb/upb.c
haberman: The allocation function takes an arbitrary pointer to user data, I think the user is only signing up to make that live until `upb_arena_free()`.

On the other hand, I am not sure anybody is using the custom allocator right now. So this is all a bit theoretical until somebody is.

One thing that is clear is that the current logic is busted if you fuse arenas that use different allocators. So rejecting anything but the global allocator is an easy short term fix that will at least prevent broken logic.

fowles: Added a restriction so that fuse only works on block allocs

haberman (approved):

Issue #305, opened 2020-07-30
Schema-independent message operations should go in upb/msg.h


haberman:
Right now https://github.com/protocolbuffers/upb/blob/master/upb/reflection.h has some schema-independent operations like:

- `upb_msg_addunknown()`
- `upb_msg_getunknown()`

These do not actually require reflection, descriptors, etc. so they should not depend on the reflection library.

They should go into `upb/msg.h` and the implementation should go in `upb/msg.c`.

For cleanliness, we should also consider moving the internal-only functions currently in `upb/msg.h` to a separate header like `upb/msg.internal.h`.

PR #381, opened 2021-04-05
Put public message interface into msg.h and moved internal functions to msg.int.h.

haberman:
Fixes: https://github.com/protocolbuffers/upb/issues/305


BUILD
```diff
@@ -59,7 +59,7 @@ cc_library(
         "upb/decode.int.h",
         "upb/encode.c",
         "upb/msg.c",
-        "upb/msg.h",
+        "upb/msg.int.h",
```
fowles, line 62: at some point we should rename `int` to `internal`.  It took me quite a while to figure out what it meant and filenames are practically free.

upb/msg.int.h
```diff
@@ -0,0 +1,603 @@
+/*
+** Our memory representation for parsing tables and messages themselves.
+** Functions in this file are used by generated code and possibly reflection.
+**
+** The definitions in this file are internal to upb.
+**/
```
fowles, line 6: every file should start with a copyright notice.  Additionally, you should create this file as a git cp of the old one so that the diff tracking and blame are better

BUILD
haberman: Sounds good, I'll follow up with a change that does this for all of the `.int.h` files.

upb/msg.int.h
haberman: I'll follow up with a change to add copyright notices to all files.

Git doesn't store rename information, it's automatically detected after the fact. See: https://stackoverflow.com/a/7941544

> Git tracks file contents, not filenames. So renaming a file without changing its content is easy for git to detect. (Git does not track, but performs detection; using git mv or git rm and git add is effectively the same.)

fowles (approved):
LGTM

PR #382, opened 2021-04-05
Added license headers to all files

haberman:
upb did not previously have per-file license headers. This PR addresses that by adding license headers to all source files (excluding auto-generated files). upb is licensed under 3-clause BSD.

I used the original years of 2009-2011 that are currently present in `LICENSE`.

I waffled a bit on whether to use C style (`/* comment */`) or C++ style (`// comment`) for C files. upb now requires C99, so C++ style comments would be technically feasible to use. I ultimately opted to go with C style comments, as most of upb still uses C style comments, and it provides a useful hint to the reader that that the headers and source files are C rather than C++.


fowles (approved):
All 2011 should be brought forward to 2021

haberman:
> All 2011 should be brought forward to 2021

Done. I also updated "Google, Inc." to "Google LLC".

fowles (approved):
LGTM

PR #390, opened 2021-04-18
Converted fasttable decoder to use __attribute__((musttail)) when it is available

haberman:
The fasttable decoder relies on the compiler generating true tail calls when dispatching from one field to the next. Without tail calls, the decoder will use stack space proportional to the number of fields parsed, which will quickly lead to stack overflow.

Prior to this PR, we relied on the fact that Clang and GCC generally do generate tail calls when optimizations are enabled. However this is not a guarantee, and it definitely won't happen in debug modes. This makes the fasttable decoder too fragile to truly roll out into production environments.

However, thanks to the `musttail` attribute just introduced in https://reviews.llvm.org/D99517, we can now get guaranteed tail calls in both debug and opt builds, as long as we are building with a new enough version of Clang. This PR adopts the code base to take advantage of the `musttail` attribute when it is availabe.

This change unfortunately requires us to turn several `__attribute__((always_inline))` functions into macros to satisfy the constraint that caller and callee for `__attribute__((musttail))` have matching function signatures. Previously we were using these `always_inline` functions as pseudo-templates, with compile-time-known parameters functioning as pseudo template parameters. Unfortunately this strategy is incompatible with `musttail`'s constraints, so these pseudo-templates must become macros. The alternative would be to use C++ and real templates instead, and while the result would certainly be more readable, abandoning upb's pure-C nature would make other things more complex (for example, the amalgamation could no longer be a single `.c`/`.h` pair).

This change is basically code size neutral. There does appear to be a slight CPU regression, and looking at assembly output I can spot some obvious inefficiencies, like push/pop in our specialized parser functions. But it appears that some regression had occurred prior to this change, and regaining this perf should probably be a separate optimization effort. Nothing in this change should fundamentally compromise the performance opportunities of this design.

```
ArenaOneAlloc                             21.7ns ± 0%  21.5ns ± 0%  -0.54%  (p=0.000 n=12+11)
ArenaInitialBlockOneAlloc                 6.34ns ± 0%  6.33ns ± 0%    ~     (p=0.099 n=11+10)
LoadDescriptor_Upb                        43.2µs ± 1%  43.3µs ± 1%    ~     (p=0.347 n=12+12)
LoadAdsDescriptor_Upb                     2.50ms ± 0%  2.49ms ± 0%  -0.52%  (p=0.000 n=11+12)
LoadDescriptor_Proto2                      240µs ± 0%   241µs ± 0%  +0.53%  (p=0.000 n=12+12)
LoadAdsDescriptor_Proto2                  12.8ms ± 0%  12.7ms ± 0%  -1.06%  (p=0.000 n=12+11)
Parse_Upb_FileDesc<UseArena,Copy>         4.85µs ± 0%  4.96µs ± 0%  +2.16%  (p=0.000 n=12+10)
Parse_Upb_FileDesc<UseArena,Alias>        3.96µs ± 0%  4.10µs ± 0%  +3.39%  (p=0.000 n=12+11)
Parse_Upb_FileDesc<InitBlock,Copy>        4.44µs ± 0%  4.57µs ± 0%  +2.92%  (p=0.000 n=9+12)
Parse_Upb_FileDesc<InitBlock,Alias>       3.61µs ± 0%  3.75µs ± 0%  +3.97%  (p=0.000 n=12+12)
Parse_Proto2<FileDesc,NoArena,Copy>       29.2µs ± 0%  29.4µs ± 0%  +0.50%  (p=0.000 n=12+12)
Parse_Proto2<FileDesc,UseArena,Copy>      20.6µs ± 2%  20.8µs ± 2%    ~     (p=0.347 n=12+12)
Parse_Proto2<FileDesc,InitBlock,Copy>     16.5µs ± 0%  16.4µs ± 0%  -0.60%  (p=0.000 n=12+12)
Parse_Proto2<FileDescSV,InitBlock,Alias>  16.6µs ± 0%  16.4µs ± 1%  -1.10%  (p=0.039 n=12+12)
SerializeDescriptor_Proto2                5.33µs ± 1%  5.38µs ± 1%  +0.99%  (p=0.000 n=12+11)
SerializeDescriptor_Upb                   12.3µs ± 0%  12.5µs ± 0%  +1.35%  (p=0.000 n=12+12)
                                                                                                                                      
                                                                                                                                      
    FILE SIZE        VM SIZE                                                                                                          
   +18%    +144  [ = ]       0    [Unmapped]                                                                                          
  -1.2%      -4  -1.2%      -4    [section .text]                                                                                     
    [NEW]      +2  [NEW]      +2    fastdecode_isdonefallback                                                                         
    [DEL]      -6  [DEL]      -6    fastdecode_longstring_noutf8                                                                      
  -0.3%    -140  -0.3%    -140    upb/decode_fast.c                                                                                   
    +5.8%     +64  +6.0%     +64    upb_pom_1bt_max64b                                                                                
    +2.7%     +64  +2.7%     +64    upb_ppv8_2bt                                                                                      
    +2.7%     +32  +2.8%     +32    upb_psm_1bt_max256b                                                                               
    +2.8%     +32  +3.0%     +32    upb_psm_1bt_max64b                                                                                
    +2.8%     +32  +3.0%     +32    upb_psm_2bt_max64b                                                                                
    +4.0%     +24  +4.2%     +24    upb_psv8_1bt                                                                                      
    +2.0%     +16  +2.1%     +16    upb_prf4_2bt                                                                                      
    +1.3%     +16  +1.4%     +16    upb_prz8_2bt                                                                                      
    -1.6%      -8  -1.7%      -8    upb_cob_1bt                                                                                       
    -0.6%     -12  -0.7%     -12    [4 Others]                                                                                        
    -2.5%     -16  -2.6%     -16    upb_pov4_1bt                                                                                      
    -1.9%     -16  -2.0%     -16    upb_ppf8_2bt                                                                                      
    -1.3%     -16  -1.3%     -16    upb_prv8_2bt                                                                                      
    -2.5%     -16  -2.7%     -16    upb_psv4_1bt                                                                                      
    -2.5%     -16  -2.6%     -16    upb_psv4_2bt                                                                                      
    -3.0%     -32  -3.1%     -32    upb_prs_2bt                                                                                       
    -2.6%     -32  -2.6%     -32    upb_prv4_2bt                                                                                      
    -4.9%     -48  -5.1%     -48    upb_prb_2bt                                                                                       
    -3.9%     -48  -4.0%     -48    upb_prv4_1bt                                                                                      
    -7.2%     -72  -7.5%     -72    upb_prb_1bt                                                                                       
    -7.8%     -88  -8.0%     -88    upb_prs_1bt                                                                                       
  [ = ]       0  -0.1%    -144    TOTAL 
```


upb/decode_fast.c
```diff
@@ -52,8 +52,9 @@
 
 #define UPB_PARSE_ARGS d, ptr, msg, table, hasbits, data
 
-#define RETURN_GENERIC(m)  \
+#define RETURN_GENERIC(m)   \
   /* fprintf(stderr, m); */ \
+  /*__builtin_trap(); */    \
```
fowles, line 57: I am confused, why does this macro expand to commented out code?

upb/decode_fast.c
```diff
@@ -66,41 +67,41 @@ typedef enum {
 UPB_NOINLINE
 static const char *fastdecode_isdonefallback(upb_decstate *d, const char *ptr,
                                              upb_msg *msg, intptr_t table,
-                                             uint64_t hasbits, int overrun) {
+                                             uint64_t hasbits, uint64_t data) {
```
fowles, line 70: should this be UPB_PARSE_PARAMS?

upb/decode_fast.c
```diff
@@ -530,126 +521,121 @@ TAGBYTES(p)
...
-      return unpacked(UPB_PARSE_ARGS);
-    } else {
-      RETURN_GENERIC("varint field tag mismatch\n");
-    }
+#define FASTDECODE_UNPACKEDFIXED(d, ptr, msg, table, hasbits, data, tagbytes, \
+                                 valbytes, card, packed)                      \
+  void *dst;                                                                  \
+  fastdecode_arr farr;                                                        \
+                                                                              \
+  if (UPB_UNLIKELY(!fastdecode_checktag(data, tagbytes))) {                   \
```
fowles, line 536: not 100% sure, but it looks like this stanza is the same in both UNPACKED macros.  Can we try and factor our the common bits into shared macros?

upb/decode_fast.c
```diff
@@ -530,126 +521,121 @@ TAGBYTES(p)
...
+
+#define FASTDECODE_PACKEDFIXED(d, ptr, msg, table, hasbits, data, tagbytes, \
+                               valbytes, unpacked)                          \
+  if (UPB_UNLIKELY(!fastdecode_checktag(data, tagbytes))) {                 \
+    if (fastdecode_flippacked(&data, tagbytes)) {                           \
+      UPB_MUSTTAIL return unpacked(UPB_PARSE_ARGS);                         \
+    } else {                                                                \
+      RETURN_GENERIC("varint field tag mismatch\n");                        \
+    }                                                                       \
+  }                                                                         \
```
fowles, line 585: similar comment here

upb/decode_internal.h
```diff
@@ -142,19 +142,22 @@ bool decode_isdone(upb_decstate *d, const char **ptr) {
   }
 }
 
+#if UPB_FASTTABLE
 UPB_INLINE
 const char *fastdecode_tagdispatch(upb_decstate *d, const char *ptr,
                                     upb_msg *msg, intptr_t table,
-                                    uint64_t hasbits, uint32_t tag) {
+                                    uint64_t hasbits, uint64_t tag) {
```
fowles, line 149: should this be the params macro?

upb/decode_internal.h
```diff
@@ -142,19 +142,22 @@ bool decode_isdone(upb_decstate *d, const char **ptr) {
...
   const upb_msglayout *table_p = decode_totablep(table);
   uint8_t mask = table;
   uint64_t data;
   size_t idx = tag & mask;
   UPB_ASSUME((idx & 7) == 0);
   idx >>= 3;
   data = table_p->fasttable[idx].field_data ^ tag;
-  return table_p->fasttable[idx].field_parser(d, ptr, msg, table, hasbits, data);
+  UPB_MUSTTAIL return table_p->fasttable[idx].field_parser(d, ptr, msg, table,
+                                                           hasbits, data);
```
fowles, line 158: ARGS macro?

gerben-stavenga (commented):
Drive by comment AWESOME!

upb/decode_fast.c
haberman: The meat of the macro is the call to `fastdecode_generic()`, these two commented-out lines are just there for convenience, to un-comment during debugging. It's the only reason that the `m` parameter to the macro exists, so we can print it out during debugging.

If you prefer I can remove the commented-out code, or add a comment to explain why it is there.

upb/decode_fast.c
haberman: Done.

upb/decode_fast.c
haberman: Done. With an extra parameter I was able to factor out the PACKED checks also.

upb/decode_fast.c
haberman: Done.

upb/decode_fast.c
fowles: a comment is fine

upb/decode_internal.h
haberman: Currently the params macro is not available in the header. I started shuffling some things around so that it would be, but it cascaded a bit and I think I'd prefer to do it in a separate change.

upb/decode_internal.h
haberman: Ditto here.

upb/decode_fast.c
haberman: Done.

haberman:
Addressed all comments, PTAL.

upb/decode_fast.c
```diff
@@ -528,93 +525,82 @@ TAGBYTES(p)
...
+                                                                               \
+  ptr += tagbytes;                                                             \
+  int size = (uint8_t)ptr[0];                                                  \
+  ptr++;                                                                       \
+  if (size & 0x80) {                                                           \
+    ptr = fastdecode_longsize(ptr, &size);                                     \
+  }                                                                            \
+                                                                               \
+  if (UPB_UNLIKELY(fastdecode_boundscheck(ptr, size, d->limit_ptr)) ||         \
+      (size % valbytes) != 0) {                                                \
```
fowles, line 581: I suspect you want the second part to be marked unlikely as well

upb/decode_fast.c
```diff
@@ -530,126 +518,110 @@ TAGBYTES(p)
...
+  dst = fastdecode_getfield(d, ptr, msg, &data, &hasbits, &farr, valbytes,     \
+                            card);                                             \
+  if (card == CARD_r) {                                                        \
+    if (UPB_UNLIKELY(!dst)) {                                                  \
+      RETURN_GENERIC("couldn't allocate array in arena\n");                    \
+    }                                                                          \
+  }                                                                            \
+                                                                               \
+  again:                                                                       \
+  if (card == CARD_r) {                                                        \
```
fowles, line 544: what is this card == CARD_r test that I see all over?

upb/decode_internal.h
fowles: SGTM

fowles (approved):

upb/decode_fast.c
haberman: `card` is the field cardinality (singular, repeated, or oneof) and `CARD_r` means "repeated field".

In this case, we only need to perform the array resize operation for repeated fields, not for singular or oneof fields.

upb/decode_fast.c
haberman: Done.

PR #396, opened 2021-05-29
Optimized decoder and paved the way for parsing extensions.

haberman:
The primary motivation for this change is to avoid referring to the `upb_msglayout` object when we are trying to fetch the `upb_msglayout` object for a sub-message. This will help pave the way for parsing extensions. We also implement several optimizations so that we can make this change without regressing performance.

Normally we compute the layout for a sub-message field like so:

```c
const upb_msglayout *get_submsg_layout(
    const upb_msglayout *layout,
    const upb_msglayout_field *field) {
  return layout->submsgs[field->submsg_index]
}
```

The reason for this indirection is to avoid storing a pointer directly in `upb_msglayout_field`, as this would double its size (from 12 to 24 bytes on 64-bit architectures) which is wasteful as this pointer is only needed for message typed fields.

However `get_submsg_layout` as written above does not work for extensions, as they will not have entries in the message's `layout->submsgs` array by nature, and we want to avoid creating an entire fake `upb_msglayout` for each such extension since that would also be wasteful.

This change removes the dependency on `upb_msglayout` by passing down the `submsgs` array instead:

```c
const upb_msglayout *get_submsg_layout(
    const upb_msglayout *const *submsgs,
    const upb_msglayout_field *field) {
  return submsgs[field->submsg_index]
}
```

This will pave the way for parsing extensions, as we can more easily create an alternative `submsgs` array for extension fields without extra overhead or waste.

Along the way several optimizations presented themselves that allow a nice increase in performance:

1. Passing the parsed `wireval` by address instead of by value ended up avoiding an expensive and useless stack copy (this is on Clang, which was used for all measurements).
2. When field numbers are densely packed, we can find a field by number with a single indexed lookup instead of linear search. At codegen time we can compute the maximum field number that will allow such an indexed lookup.
3. For fields that do require linear search, we can start the linear search at the location where we found the previous field, taking advantage of the fact that field numbers are generally increasing.
4. When the hasbit index is less than 32 (the common case) we can use a less expensive code sequence to set it.
5. We check for the hasbit case before the oneof case, as optional fields are more common than oneof fields.

Benchmark results indicate a 20-25% reduction in parse time with a small code size increase:

```
ArenaOneAlloc                             21.3ns ± 0%  21.9ns ± 1%   +3.08%  (p=0.000 n=11+12)
ArenaInitialBlockOneAlloc                 6.32ns ± 0%  6.32ns ± 0%     ~     (p=0.219 n=12+12)
LoadDescriptor_Upb                        53.5µs ± 1%  51.2µs ± 2%   -4.30%  (p=0.000 n=12+12)
LoadAdsDescriptor_Upb                     2.75ms ± 0%  2.67ms ± 2%   -2.68%  (p=0.000 n=10+10)
LoadDescriptor_Proto2                      240µs ± 0%   239µs ± 0%   -0.22%  (p=0.000 n=12+12)
LoadAdsDescriptor_Proto2                  12.6ms ± 0%  12.8ms ± 0%   +1.17%  (p=0.000 n=12+11)
Parse_Upb_FileDesc<UseArena,Copy>         13.2µs ± 0%  10.1µs ± 0%  -23.85%  (p=0.000 n=12+12)
Parse_Upb_FileDesc<UseArena,Alias>        11.5µs ± 0%   8.8µs ± 0%  -22.92%  (p=0.000 n=11+11)
Parse_Upb_FileDesc<InitBlock,Copy>        12.7µs ± 0%   9.6µs ± 0%  -24.68%  (p=0.000 n=11+12)
Parse_Upb_FileDesc<InitBlock,Alias>       11.0µs ± 0%   8.5µs ± 0%  -22.89%  (p=0.000 n=12+11)
Parse_Proto2<FileDesc,NoArena,Copy>       29.3µs ± 0%  29.4µs ± 0%   +0.30%  (p=0.001 n=12+12)
Parse_Proto2<FileDesc,UseArena,Copy>      20.4µs ± 2%  21.1µs ± 2%   +3.38%  (p=0.000 n=12+12)
Parse_Proto2<FileDesc,InitBlock,Copy>     16.5µs ± 0%  17.5µs ± 0%   +5.98%  (p=0.000 n=12+12)
Parse_Proto2<FileDescSV,InitBlock,Alias>  16.2µs ± 0%  17.4µs ± 0%   +7.52%  (p=0.000 n=12+11)
SerializeDescriptor_Proto2                5.35µs ± 1%  5.47µs ± 1%   +2.19%  (p=0.000 n=12+12)
SerializeDescriptor_Upb                   12.9µs ± 0%  13.1µs ± 0%   +1.41%  (p=0.000 n=12+11)


    FILE SIZE        VM SIZE    
  +0.6%     +64  +0.6%     +64    upb/decode.c
    +0.7%     +64  +0.7%     +64    decode_msg
  +0.3%     +48  +0.3%     +48    upb/def.c
    +1.1%     +48  +1.1%     +48    _upb_symtab_addfile
  -3.6%    -112  [ = ]       0    [Unmapped]
  [ = ]       0  +0.1%    +112    TOTAL
```

### DIFF
--- a/upb/decode.c
+++ b/upb/decode.c
@@ -299,24 +299,42 @@ static void decode_munge(int type, wireval *val) {
 }
 
 static const upb_msglayout_field *upb_find_field(const upb_msglayout *l,
-                                                 uint32_t field_number) {
+                                                 uint32_t field_number,
+                                                 int *last_field_index) {
   static upb_msglayout_field none = {0, 0, 0, 0, 0, 0};
 
-  /* Lots of optimization opportunities here. */
-  int i;
   if (l == NULL) return &none;
-  for (i = 0; i < l->field_count; i++) {
-    if (l->fields[i].number == field_number) {
-      return &l->fields[i];
+
+  size_t idx = (size_t)field_number - 1;  // 0 wraps to SIZE_MAX
+  if (idx < l->dense_below) {
+    goto found;
+  }
+
+  int last = *last_field_index;
+  for (idx = last; idx < l->field_count; idx++) {
+    if (l->fields[idx].number == field_number) {
+      goto found;
+    }
+  }
+
+  for (idx = 0; idx < last; idx++) {
+    if (l->fields[idx].number == field_number) {
+      goto found;
     }
   }
 
   return &none; /* Unknown field. */
+
+ found:
+  UPB_ASSERT(l->fields[idx].number == field_number);
+  *last_field_index = idx;
+  return &l->fields[idx];
 }
 
-static upb_msg *decode_newsubmsg(upb_decstate *d, const upb_msglayout *layout,
+static upb_msg *decode_newsubmsg(upb_decstate *d,
+                                 const upb_msglayout *const *submsgs,
                                  const upb_msglayout_field *field) {
-  const upb_msglayout *subl = layout->submsgs[field->submsg_index];
+  const upb_msglayout *subl = submsgs[field->submsg_index];
   return _upb_msg_new_inl(subl, &d->arena);
 }
 
@@ -346,9 +364,10 @@ static const char *decode_readstr(upb_decstate *d, const char *ptr, int size,
 
 UPB_FORCEINLINE
 static const char *decode_tosubmsg(upb_decstate *d, const char *ptr,
-                                   upb_msg *submsg, const upb_msglayout *layout,
+                                   upb_msg *submsg, 
+                                   const upb_msglayout *const *submsgs,
                                    const upb_msglayout_field *field, int size) {
-  const upb_msglayout *subl = layout->submsgs[field->submsg_index];
+  const upb_msglayout *subl = submsgs[field->submsg_index];
   int saved_delta = decode_pushlimit(d, ptr, size);
   if (--d->depth < 0) decode_err(d);
   if (!decode_isdone(d, &ptr)) {
@@ -377,15 +396,17 @@ static const char *decode_group(upb_decstate *d, const char *ptr,
 
 UPB_FORCEINLINE
 static const char *decode_togroup(upb_decstate *d, const char *ptr,
-                                  upb_msg *submsg, const upb_msglayout *layout,
+                                  upb_msg *submsg,
+                                  const upb_msglayout *const *submsgs,
                                   const upb_msglayout_field *field) {
-  const upb_msglayout *subl = layout->submsgs[field->submsg_index];
+  const upb_msglayout *subl = submsgs[field->submsg_index];
   return decode_group(d, ptr, submsg, subl, field->number);
 }
 
 static const char *decode_toarray(upb_decstate *d, const char *ptr,
-                                  upb_msg *msg, const upb_msglayout *layout,
-                                  const upb_msglayout_field *field, wireval val,
+                                  upb_msg *msg,
+                                  const upb_msglayout *const *submsgs,
+                                  const upb_msglayout_field *field, wireval *val,
                                   int op) {
   upb_array **arrp = UPB_PTR_AT(msg, field->offset, void);
   upb_array *arr = *arrp;
@@ -407,27 +428,27 @@ static const char *decode_toarray(upb_decstate *d, const char *ptr,
       /* Append scalar value. */
       mem = UPB_PTR_AT(_upb_array_ptr(arr), arr->len << op, void);
       arr->len++;
-      memcpy(mem, &val, 1 << op);
+      memcpy(mem, val, 1 << op);
       return ptr;
     case OP_STRING:
-      decode_verifyutf8(d, ptr, val.size);
+      decode_verifyutf8(d, ptr, val->size);
       /* Fallthrough. */
     case OP_BYTES: {
       /* Append bytes. */
       upb_strview *str = (upb_strview*)_upb_array_ptr(arr) + arr->len;
       arr->len++;
-      return decode_readstr(d, ptr, val.size, str);
+      return decode_readstr(d, ptr, val->size, str);
     }
     case OP_SUBMSG: {
       /* Append submessage / group. */
-      upb_msg *submsg = decode_newsubmsg(d, layout, field);
+      upb_msg *submsg = decode_newsubmsg(d, submsgs, field);
       *UPB_PTR_AT(_upb_array_ptr(arr), arr->len * sizeof(void *), upb_msg *) =
           submsg;
       arr->len++;
       if (UPB_UNLIKELY(field->descriptortype == UPB_DTYPE_GROUP)) {
-        return decode_togroup(d, ptr, submsg, layout, field);
+        return decode_togroup(d, ptr, submsg, submsgs, field);
       } else {
-        return decode_tosubmsg(d, ptr, submsg, layout, field, val.size);
+        return decode_tosubmsg(d, ptr, submsg, submsgs, field, val->size);
       }
     }
     case OP_FIXPCK_LG2(2):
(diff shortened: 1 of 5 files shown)


upb/def.c
```diff
@@ -1018,14 +1018,20 @@ static int field_number_cmp(const void *p1, const void *p2) {
...
+  int dense_below = 0;
   for (i = 0; i < n; i++) {
     upb_fielddef *f = (upb_fielddef*)upb_msgdef_itof(m, fields[i].number);
     UPB_ASSERT(f);
     f->layout_index = i;
+    if (upb_fielddef_number(f) == i + 1) {
+      dense_below = upb_fielddef_number(f);
+    }
   }
+  l->dense_below = dense_below;
```
fowles, line 1034: As a potential heuristic:

- once I have k gaps within N of each other, I will rollback to the last gap and stop being dense

so if you say 3 gaps within 8 of each other, your worst case loss to gaps is 1/4.

upb/msg_internal.h
```diff
@@ -65,6 +65,7 @@ struct upb_msglayout {
   uint16_t size;
   uint16_t field_count;
   bool extendable;
+  uint8_t dense_below;
```
fowles, line 68: dense_below maxes at 256, but I don't see any test for that in your build above

upb/msg_internal.h
```diff
@@ -131,7 +132,11 @@ UPB_INLINE bool _upb_hasbit(const upb_msg *msg, size_t idx) {
 }
 
 UPB_INLINE void _upb_sethas(const upb_msg *msg, size_t idx) {
-  (*UPB_PTR_AT(msg, idx / 8, char)) |= (char)(1 << (idx % 8));
+  if (UPB_LIKELY(idx < 32)) {
+    (*UPB_PTR_AT(msg, 0, uint32_t)) |= (1UL << idx);
+  } else {
+    (*UPB_PTR_AT(msg, idx / 8, char)) |= (char)(1 << (idx % 8));
+  }
```
fowles, line 139: I am skeptical of this micro optimization. I think it would be better to pull this into a separate change.  micro optimizations like this often look good in benchmarks but introduce more code which then polutes your icache in real world systems

upb/decode.c
```diff
@@ -299,24 +299,42 @@ static void decode_munge(int type, wireval *val) {
...
   static upb_msglayout_field none = {0, 0, 0, 0, 0, 0};
 
-  /* Lots of optimization opportunities here. */
-  int i;
   if (l == NULL) return &none;
-  for (i = 0; i < l->field_count; i++) {
-    if (l->fields[i].number == field_number) {
-      return &l->fields[i];
+
+  size_t idx = (size_t)field_number - 1;  // 0 wraps to SIZE_MAX
```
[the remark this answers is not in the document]
haberman: The cast binds tighter: https://en.cppreference.com/w/c/language/operator_precedence

I can add parens.

upb/decode.c
```diff
@@ -299,24 +299,42 @@ static void decode_munge(int type, wireval *val) {
...
+  if (idx < l->dense_below) {
+    goto found;
+  }
+
+  int last = *last_field_index;
```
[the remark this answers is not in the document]
haberman: Done.

upb/decode.c
```diff
@@ -299,24 +299,42 @@ static void decode_munge(int type, wireval *val) {
...
+  for (idx = 0; idx < last; idx++) {
+    if (l->fields[idx].number == field_number) {
+      goto found;
     }
   }
 
   return &none; /* Unknown field. */
+
+ found:
+  UPB_ASSERT(l->fields[idx].number == field_number);
```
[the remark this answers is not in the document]
haberman: Agreed it's a good opportunity, but I'd rather punt on it for now since it is a more "tunable" optimization that would require some study/heuristics to decide on good values for the knobs.

upb/decode.c
```diff
@@ -299,24 +299,42 @@ static void decode_munge(int type, wireval *val) {
...
+  *last_field_index = idx;
+  return &l->fields[idx];
 }
 
-static upb_msg *decode_newsubmsg(upb_decstate *d, const upb_msglayout *layout,
+static upb_msg *decode_newsubmsg(upb_decstate *d,
+                                 const upb_msglayout *const *submsgs,
```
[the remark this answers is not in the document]
haberman: Done. (I hadn't heard it called "east cost" before. :) )

upb/def.c
haberman: Yes seems possibly promising. Though some users may wish to always prioritize code size. Maybe less size-conscious users are better served with fast tables? It seems like promising future work.

upb/msg_internal.h
haberman: Good catch, I checked this in the C++ codegen but not in the dynamic table building. Done.

upb/msg_internal.h
haberman: `__builtin_expect()` (aka `UPB_LIKELY()`) will make the likely case into straight-line code, and put the unlikely case far away in memory that will likely never hit the cache unless the unexpected case is hit. https://godbolt.org/z/P1fodxfxd

Also the size profiling results show that this particular optimization only grows the code by 16 or 32 bytes total (this optimization was pulled into `upb_msg_set()` and `upb_msg_mutable()`.

I can move this to a separate change, but I think it is a well-grounded optimization.

haberman: I would add though that I share your concern about icache in general, and upb is very size-sensitive in general, so I appreciate the attention to this. :)

haberman: Benchmarks at the moment seem to be looking better without it, so it looks like it was a good call, thanks.

upb/def.c
fowles: agreed, future work

upb/def.c
```diff
@@ -1018,14 +1018,20 @@ static int field_number_cmp(const void *p1, const void *p2) {
...
+static void assign_layout_indices(const upb_msgdef *m, upb_msglayout *l,
+                                  upb_msglayout_field *fields) {
   int i;
   int n = upb_msgdef_numfields(m);
+  int dense_below = 0;
   for (i = 0; i < n; i++) {
     upb_fielddef *f = (upb_fielddef*)upb_msgdef_itof(m, fields[i].number);
     UPB_ASSERT(f);
     f->layout_index = i;
+    if (i <= 254 && upb_fielddef_number(f) == i + 1) {
```
fowles, line 1030: this uses 254, but the version below using uint8_max which is 256.  Which should it be?

upb/msg_internal.h
fowles: the / and the % are likely a single instruction.  Even though they are an expensive instruction, it is still just one

upb/def.c
haberman: `UINT8_MAX` is 255 (0xff). If the predicate is true, we will assign `i + 1`, so `i <= 254` is the correct check I believe.

I wrote this predicate slightly differently from the one in the code generator because this one is size/speed-sensitive. I did the range check against `i` instead of the field number, because `i` is the loop index, which makes it easier for the compiler to reason about, and if `i > 254` we can avoid the load of `f->number` altogether.

haberman: I changed the logic to use `UINT8_MAX` as the constant we compare against.

fowles (approved):

haberman:
Bazel errors are unrelated, see: https://github.com/bazelbuild/bazel/issues/13526#issuecomment-852229300

PR #408, opened 2021-08-14
Remove CMake support.

haberman:
I don't believe anyone is currently using upb's CMake support. Removing this will simplify the upb repo and reduce maintenance burden.

Removing CMake support also ties in with upb's recent shift/clarification that upb [is not generally offered as a C library for direct consumption](https://github.com/protocolbuffers/upb#%CE%BCpb-small-fast-c-protos). Users are still free to use upb's C API if desired, as long as you can tolerate API/ABI breaks, but we won't generally be maintaining integrations specifically aimed at making this easier (like CMake support).

A few options for anybody who is currently using upb's CMake support:
1. Depend on upb using Bazel.
2. Build upb into an amalgamation using any of the `upb_amalgamation()` builds in the Bazel `BUILD` file. This will build upb into a single `.c` file that doesn't need any build system.


coryan:
> I don't believe anyone is currently using upb's CMake support. Removing this will simplify the upb repo and reduce maintenance burden.

Not quite.  The vcpkg port for upb does, with a patch to support `install` targets:

https://github.com/microsoft/vcpkg/blob/master/ports/upb/add-cmake-install.patch

The library is used as a dependency of gRPC, and therefore any project that depends on gRPC via `vcpkg`.

/FYI: @devjgm

AnticliMaxtic:
We are investigating extending and using the cmake support for upb so would appreciate it if it is not axed just yet...

calvin2021y:
please don't remove it.

PR #414, opened 2021-08-18
Beginning of Python+upb: DescriptorPool and a few Descriptor types

haberman:
This PR puts in place the structure and an initial set of objects for the Python upb extension.

This PR adds support for constructing `DescriptorPool`, adding serialized descriptors, and getting a few types of descriptors.

Since the extension can't pass any of the existing unit tests yet, a temporary unit test is added to exercise what it can do.

Subsequent PRs will flesh this out, adding support for more descriptor types and for messages themselves.


haberman:
Test is failing under ASAN because ASAN is not compiled into the Python interpreter.

I'll work on figuring out how to disable the test in this configuration. But meanwhile, the code is ready to review.

esorot (commented):
soft LGTM: reviewed on high level. Need to spend some more time to fit the picture together.

python/descriptor.c
```diff
@@ -0,0 +1,247 @@
...
+static PyObject *PyUpb_DescriptorBase_NewInternal(PyTypeObject *type,
+                                                  const void *def,
+                                                  PyObject *pool) {
+  PyUpb_DescriptorBase *base = (PyUpb_DescriptorBase *)PyUpb_ObjCache_Get(def);
+
+  if (!base) {
+    base = PyObject_New(PyUpb_DescriptorBase, type);
+    base->pool = pool;
+    base->def = def;
+    Py_INCREF(pool);
```
fowles, line 65: you use Py_INCREF here indicating it is not null and Py_CLEAR below.  feels like one of these is incosnstient

python/descriptor_pool.c
```diff
@@ -0,0 +1,216 @@
...
+                                               PyObject* db) {
+  PyUpb_DescriptorPool* pool = PyObject_New(PyUpb_DescriptorPool, type);
+  pool->symtab = upb_symtab_new();
+  pool->db = db;
+  return &pool->ob_base;
+}
+
+static void PyUpb_DescriptorPool_Dealloc(PyUpb_DescriptorPool* self) {
+  upb_symtab_free(self->symtab);
+  Py_CLEAR(self->db);
```
fowles, line 54: where is the corresponding INC for this?

python/descriptor_pool.c
```diff
@@ -0,0 +1,216 @@
...
+
+  upb_status status;
+  upb_status_clear(&status);
+
+  const upb_filedef* filedef = upb_symtab_addfile(self->symtab, proto, &status);
+  if (!filedef) {
+    PyErr_Format(PyExc_TypeError,
+                 "Couldn't build proto file into descriptor pool: %s",
+                 upb_status_errmsg(&status));
+    return NULL;
```
fowles, line 119: I think you are msising the arena free for all the error cases

python/descriptor_pool.c
```diff
@@ -0,0 +1,216 @@
...
+
+// -----------------------------------------------------------------------------
+// Top Level
+// -----------------------------------------------------------------------------
+
+bool PyUpb_InitDescriptorPool(PyObject* m) {
+  PyTypeObject* descriptor_pool_type =
+      AddObject(m, "DescriptorPool", &PyUpb_DescriptorPool_Spec);
+
+  if (!descriptor_pool_type) return false;
```
fowles, line 210: you are inconsistent within this file if you do the 1 line or 3 line if for an early return like this.  I prefer 1, but don't really care as long as you are consistent

python/minimal_test.py
```diff
@@ -0,0 +1,55 @@
...
+# DISCLAIMED. IN NO EVENT SHALL Google LLC BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
+"""A bare-bones unit test, to be removed once upb can pass existing unit tests."""
```
fowles, line 27: not sure removing it is right.  There is value to minimal tests

python/protobuf.c
```diff
@@ -0,0 +1,128 @@
...
+  upb_inttable_remove(&s->obj_cache, (uintptr_t)key, &val);
+  assert(upb_value_getptr(val));
+}
+
+PyObject *PyUpb_ObjCache_Get(const void *key) {
+  PyUpb_ModuleState *s = PyUpb_ModuleState_Get();
+  upb_value val;
+  if (upb_inttable_lookup(&s->obj_cache, (uintptr_t)key, &val)) {
+    PyObject *ret = upb_value_getptr(val);
+    Py_INCREF(ret);
```
fowles, line 79: where is the corresponding DECREF?

python/protobuf.h
```diff
@@ -0,0 +1,93 @@
...
+#define PYUPB_MODULE_NAME "google.protobuf.pyext._message"
+
+// -----------------------------------------------------------------------------
+// ModuleState
+// -----------------------------------------------------------------------------
+
+// We store all "global" state in this struct instead of using (C) global
+// variables. This makes this extension compatible with sub-interpreters.
+
+typedef struct {
```
fowles, line 53: do you need one of thos magic Py_HEAD things?

python/descriptor.c
haberman: Done.

python/descriptor_pool.c
haberman: Done. This made me realize this type does need GC, since the database is a user object that could point back to the pool.

python/descriptor_pool.c
haberman: Done. (I had a `done` label but forgot to use it).

python/descriptor_pool.c
haberman: Done.

python/minimal_test.py
haberman: Done.

python/protobuf.c
haberman: This is a function that returns a reference to the caller when it succeeds. I added a comment describing this to the header file.

python/protobuf.h
haberman: Not here, because this is not a Python object. It is a "module state", which is just a formless hunk of memory given to the module to store arbitrary state. Note that it returns `void*`, not `PyObject*`: https://docs.python.org/3/c-api/module.html#c.PyModule_GetState

fowles (approved):

anandolee (approved):

PR #416, opened 2021-08-28
Implemented upb_enumvaldef, for storing information about enumvals.

haberman:
Previously `upb_enumdef` (the upb equivalent of `proto2::EnumDescriptor`) just contained `name -> number` and `number -> name` maps. There was no explicit def for each individual enum value.

We need `upb_enumvaldef` for two reasons:

1. (immediate need): we need to be able to look up all enum values underneath a given message. Thanks to proto's weird scoping rules, these enum values are *not* in the scope of the enum itself, so the `upb_enumdef`'s table cannot be used to perform this lookup. We unfortunately have to look up the enum values in the `upb_symtab` global map, since the value could come from any number of enums under the message.
2. (future need): Any enum value can have custom options attached, and we need a place to store these custom options.

There is minor code size growth, which is fine, but unfortunately this regresses the Ads benchmark for schema loading by 10%.

```
name                                      old time/op  new time/op  delta                                                                                                     
ArenaOneAlloc                             20.9ns ± 0%  21.4ns ± 1%   +2.25%  (p=0.000 n=12+12)           
ArenaInitialBlockOneAlloc                 6.29ns ± 0%  6.28ns ± 0%   -0.05%  (p=0.000 n=12+11)           
LoadDescriptor_Upb                        53.6µs ± 1%  54.8µs ± 1%   +2.24%  (p=0.000 n=12+12)           
LoadAdsDescriptor_Upb                     3.13ms ± 0%  3.45ms ± 0%  +10.22%  (p=0.000 n=12+11)           
LoadDescriptor_Proto2                      240µs ± 0%   239µs ± 0%   -0.48%  (p=0.000 n=12+12)           
LoadAdsDescriptor_Proto2                  14.5ms ± 0%  14.4ms ± 0%   -0.63%  (p=0.000 n=11+10)           
Parse_Upb_FileDesc<UseArena,Copy>         10.8µs ± 0%  11.1µs ± 0%   +3.24%  (p=0.000 n=12+11)           
Parse_Upb_FileDesc<UseArena,Alias>        9.00µs ± 0%  9.35µs ± 0%   +3.87%  (p=0.000 n=12+12)                                                                                
Parse_Upb_FileDesc<InitBlock,Copy>        10.3µs ± 0%  10.7µs ± 0%   +3.39%  (p=0.000 n=12+12)           
Parse_Upb_FileDesc<InitBlock,Alias>       8.67µs ± 0%  8.87µs ± 0%   +2.29%  (p=0.000 n=12+12)           
[...]

PR #417, opened 2021-08-29
Adds upb_strtable_tryinsert() to increase the efficiency of descriptor loading.

haberman:
This PR adds a `upb_strtable_tryinsert()` function to `upb_strtable`, to avoid a double lookup in cases where we are wanting to insert but detect duplicates. The existing function `upb_strtable_insert()` requires as a precondition that the key does not exist in the table, forcing a lookup followed by insert. `upb_strtable_tryinsert()` will return a status code if the key already existed.

This also removes some deprecated functions in `upb/def.h`. I don't expect that gRPC is using these (they only use reflection for parsing and serializing JSON, not to actually inspect the schema), so I don't expect this to be an intrusive change.

This PR regains most but not all of the performance lost in https://github.com/protocolbuffers/upb/pull/416 when we load descriptors.

```
ArenaOneAlloc                             21.4ns ± 0%  20.9ns ± 0%   -2.55%  (p=0.000 n=12+12)
ArenaInitialBlockOneAlloc                 6.28ns ± 0%  6.94ns ± 0%  +10.46%  (p=0.000 n=12+12)
LoadDescriptor_Upb                        54.8µs ± 1%  48.8µs ± 1%  -10.91%  (p=0.000 n=12+12)
LoadAdsDescriptor_Upb                     3.49ms ± 3%  3.23ms ± 0%   -7.56%  (p=0.000 n=12+11)
LoadDescriptor_Proto2                      239µs ± 0%   239µs ± 0%   +0.08%  (p=0.028 n=12+12)
LoadAdsDescriptor_Proto2                  14.7ms ± 0%  14.7ms ± 0%   +0.19%  (p=0.000 n=10+10)
Parse_Upb_FileDesc<UseArena,Copy>         11.0µs ± 0%  10.1µs ± 0%   -8.63%  (p=0.000 n=11+12)
Parse_Upb_FileDesc<UseArena,Alias>        9.37µs ± 0%  8.65µs ± 0%   -7.75%  (p=0.000 n=11+12)
Parse_Upb_FileDesc<InitBlock,Copy>        10.5µs ± 0%   9.6µs ± 0%   -8.57%  (p=0.000 n=12+12)
Parse_Upb_FileDesc<InitBlock,Alias>       8.87µs ± 0%  8.30µs ± 0%   -6.44%  (p=0.000 n=12+12)
Parse_Proto2<FileDesc,NoArena,Copy>       29.6µs ± 0%  29.4µs ± 0%   -0.63%  (p=0.000 n=12+12)
Parse_Proto2<FileDesc,UseArena,Copy>      20.9µs ± 2%  20.8µs ± 3%     ~     (p=0.266 n=12+12)
Parse_Proto2<FileDesc,InitBlock,Copy>     16.7µs ± 0%  16.7µs ± 0%     ~     (p=0.143 n=12+12)
Parse_Proto2<FileDescSV,InitBlock,Alias>  16.8µs ± 0%  16.8µs ± 0%   -0.10%  (p=0.003 n=10+10)
SerializeDescriptor_Proto2                5.33µs ± 1%  5.36µs ± 1%     ~     (p=0.219 n=12+12)
SerializeDescriptor_Upb                   12.4µs ± 0%  12.5µs ± 0%   +0.26%  (p=0.000 n=12+12)


    FILE SIZE        VM SIZE    
  +7.3%    +742  +7.6%    +696    upb/table.c
    [NEW]    +743  [NEW]    +696    upb_strtable_tryinsert
    +6.1%      +7  [ = ]       0    upb_inttable_init
    -0.2%      -8  [ = ]       0    upb_inttable_compact
  +0.1%      +1  [ = ]       0    [section .strtab]
  -1.8%    -336  -2.1%    -336    upb/def.c
    -2.2%     -64  -2.3%     -64    create_msgdef
    -7.1%     -96  -7.3%     -96    create_enumdef
    -9.1%    -176  -9.3%    -176    create_fielddef
 -13.7%    -359  [ = ]       0    [Unmapped]
  +0.0%     +48  +0.3%    +360    TOTAL
```

I can't explain why some unrelated benchmarks (like arena and parse) are showing wide swings. It seems to be a benchmarking artifact, possible related to alignment.

### DIFF
--- a/upb/table.c
+++ b/upb/table.c
@@ -194,14 +194,15 @@ static bool lookup(const upb_table *t, lookupkey_t key, upb_value *v,
   }
 }
 
-/* The given key must not already exist in the table. */
-static void insert(upb_table *t, lookupkey_t key, upb_tabkey tabkey,
+static bool insert(upb_table *t, lookupkey_t key, upb_tabkey tabkey,
                    upb_value val, uint32_t hash,
-                   hashfunc_t *hashfunc, eqlfunc_t *eql) {
+                   hashfunc_t *hashfunc, eqlfunc_t *eql, bool allow_failure) {
   upb_tabent *mainpos_e;
   upb_tabent *our_e;
 
-  UPB_ASSERT(findentry(t, key, hash, eql) == NULL);
+  if (!allow_failure) {
+    UPB_ASSERT(findentry(t, key, hash, eql) == NULL);
+  }
 
   t->count++;
   mainpos_e = getentry_mutable(t, hash);
@@ -217,7 +218,16 @@ static void insert(upb_table *t, lookupkey_t key, upb_tabkey tabkey,
     upb_tabent *chain = getentry_mutable(t, hashfunc(mainpos_e->key));
     if (chain == mainpos_e) {
       /* Existing ent is in its main position (it has the same hash as us, and
-       * is the head of our chain).  Insert to new ent and append to this chain. */
+       * is the head of our chain). */
+      if (allow_failure) {
+        const upb_tabent *e = mainpos_e;
+        while (1) {
+          if (eql(e->key, key)) return false;  // Already present?
+          if ((e = e->next) == NULL) break;
+        }
+      }
+
+      /* Insert to new ent and append to this chain. */
       new_e->next = mainpos_e->next;
       mainpos_e->next = new_e;
       our_e = new_e;
@@ -238,6 +248,7 @@ static void insert(upb_table *t, lookupkey_t key, upb_tabkey tabkey,
   our_e->key = tabkey;
   our_e->val.val = val.val;
   UPB_ASSERT(findentry(t, key, hash, eql) == our_e);
+  return true;
 }
 
 static bool rm(upb_table *t, lookupkey_t key, upb_value *val,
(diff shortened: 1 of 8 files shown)


upb/table.c
```diff
@@ -506,10 +517,35 @@ bool upb_strtable_insert(upb_strtable *t, const char *k, size_t len,
...
+  insert(&t->t, key, tabkey, v, hash, &strhash, &streql, false);
   return true;
 }
 
+upb_tryinsert_status upb_strtable_tryinsert(upb_strtable *t, const char *k,
+                                            upb_value val, upb_arena *a) {
+  lookupkey_t key;
+  upb_tabkey tabkey;
+  uint32_t hash;
+  size_t len = strlen(k);
```
fowles, line 529: might as well push the strlen down below to the point where you need it

upb/table.c
```diff
@@ -506,10 +517,35 @@ bool upb_strtable_insert(upb_strtable *t, const char *k, size_t len,
...
+
+  if (isfull(&t->t)) {
```
fowles, line 531: this will grow the table even if the key is already present which seems undesirable

upb/table.c
```diff
@@ -194,14 +194,15 @@ static bool lookup(const upb_table *t, lookupkey_t key, upb_value *v,
   }
 }
 
-/* The given key must not already exist in the table. */
-static void insert(upb_table *t, lookupkey_t key, upb_tabkey tabkey,
+static bool insert(upb_table *t, lookupkey_t key, upb_tabkey tabkey,
                    upb_value val, uint32_t hash,
-                   hashfunc_t *hashfunc, eqlfunc_t *eql) {
+                   hashfunc_t *hashfunc, eqlfunc_t *eql, bool allow_failure) {
```
fowles, line 199: instead of an allow_failure bool, you should make this method be an unconditional try_insert and then have callers assert whichever way they want

PR #428, opened 2021-10-09
Optimize varint encoding

gerben-stavenga:
1) Rework the varint encoding loop to be less code
https://godbolt.org/z/r9aW5hh1e

2) Use fixed size power of 2 memcpy that compiles to a single load/store instead of a function call.

Didn't do extensive benchmarks

New
BM_SerializeDescriptor_Upb                          9009 ns         9009 ns        77066 bytes_per_second=794.783M/s

Old
BM_SerializeDescriptor_Upb                          9416 ns         9416 ns        70574 bytes_per_second=760.212M/s

### DIFF
--- a/upb/encode.c
+++ b/upb/encode.c
@@ -42,11 +42,11 @@
 
 static char *encode_varint64(uint64_t val, char *buf)
 {
-  do
+  while (val > 127)
   {
     *buf++ = val | 0x80;
     val >>= 7;
-  } while (val > 127);
+  }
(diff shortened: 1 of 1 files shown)


haberman:
Sorry I didn't merge this before, I had some similar work in progress and I wanted to avoid merge conflicts.

I didn't end up making ptr local because I didn't see a big improvement. But I might have been doing it wrong.

If you want to fix the merge conflicts and re-send, please feel free.

PR #429, opened 2021-10-12
Switched to utf8_range for UTF8 verification

haberman:
The primary motivation here is to fix a bug: the old UTF8 verification was not correctly rejecting surrogate code points.

However another good reason for this change is performance. upb currently has no benchmarks for this (we should add one at some point) but benchmarks posted at https://github.com/cyb70289/utf8 suggest that this should be an O(10x) speedup for UTF8 verification.

There is a smalll amount of code size growth:

```
    FILE SIZE        VM SIZE    
  [NEW] +1.18Ki  [NEW] +1.09Ki    third_party/utf8_range/utf8_range.c
    [NEW]    +807  [NEW]    +766    utf8_range2
    [NEW]    +386  [NEW]    +336    utf8_naive
    [NEW]      +8  [NEW]      +8    __libc_csu_fini
    [NEW]      +8  [NEW]      +8    __libc_csu_init
  +0.0%     +10  [ = ]       0    upb/def.c
    +7.0%     +10  [ = ]       0    upb_fielddef_isstring
  -0.0%      -1  -0.0%      -1    upb/encode.c
    +4.6%     +15  +5.2%     +15    encode_mapentry
    [DEL]      -8  [DEL]      -8    __libc_csu_fini
    [DEL]      -8  [DEL]      -8    __libc_csu_init
  -0.3%      -2  [ = ]       0    [section .strtab]
    +1.3%      +5  [ = ]       0    [section .strtab]
   -20.6%      -7  [ = ]       0    _GLOBAL_OFFSET_TABLE_
  -0.1%      -5  [ = ]       0    tests/conformance_upb.c
    +9.3%      +8  [ = ]       0    stderr
    -6.0%     -13  [ = ]       0    main
  -0.1%     -11  [ = ]       0    upb/table.c
    -0.2%      -1  [ = ]       0    upb_inttable_sizedinit
    -5.3%     -10  [ = ]       0    upb_inttable_done
  -3.7%     -13  -3.7%     -13    [section .text]
    [NEW]      +2  [NEW]      +2    utf8_range2
    [DEL]     -15  [DEL]     -15    encode_mapentry
  -0.2%     -19  [ = ]       0    bazel-out/k8-opt/bin/external/com_google_protobuf/google/protobuf/descriptor.upb.c
    +8.4%      +8  [ = ]       0    google_protobuf_ServiceDescriptorProto_msginit
    +8.0%      +7  [ = ]       0    google_protobuf_SourceCodeInfo_msginit
    +6.2%      +6  [ = ]       0    google_protobuf_SourceCodeInfo_Location_msginit
    -5.2%      -5  [ = ]       0    google_protobuf_FieldOptions_CType_enuminit
    -5.1%      -6  [ = ]       0    google_protobuf_MethodOptions_IdempotencyLevel_enuminit
    -7.4%      -7  [ = ]       0    google_protobuf_DescriptorProto_msginit
    -7.0%      -7  [ = ]       0    google_protobuf_FieldOptions_JSType_enuminit
    -6.9%      -7  [ = ]       0    google_protobuf_MethodDescriptorProto_msginit
    -7.9%      -8  [ = ]       0    google_protobuf_OneofDescriptorProto_msginit
  -3.1%    -274  -2.9%    -240    upb/decode.c
    +0.3%     +16  +0.3%     +16    decode_msg
    +2.9%      +7  [ = ]       0    decode_isdonefallback
    [DEL]    -297  [DEL]    -256    upb_utf8_offsets
 -25.6%    -862  [ = ]       0    [Unmapped]
  +0.0%     +32  +0.6%    +864    TOTAL
```

### DIFF
--- a/upb/decode_internal.h
+++ b/upb/decode_internal.h
@@ -78,18 +78,20 @@ bool decode_verifyutf8_inl(const char *ptr, int len) {
   while (end - ptr >= 8) {
     uint64_t data;
     memcpy(&data, ptr, 8);
-    if (data & 0x8080808080808080) break;
+    if (data & 0x8080808080808080) goto non_ascii;
     ptr += 8;
   }
 
   // Check one byte at a time for non-ASCII.
   while (ptr < end) {
-    if (*ptr & 0x80) break;
+    if (*ptr & 0x80) goto non_ascii;
     ptr++;
   }
 
-  return ptr == end ? true
-                    : utf8_range2((const unsigned char *)ptr, end - ptr) == 0;
+  return true;
+
+ non_ascii:
+  return utf8_range2((const unsigned char *)ptr, end - ptr) == 0;
 }
 
 /* x86-64 pointers always have the high 16 bits matching. So we can shift


third_party/utf8_range/utf8_range.h
```diff
@@ -0,0 +1,9 @@
+
+#if defined(__ARM_NEON) || defined(__SSE4_1__)
+int utf8_range2(const unsigned char *data, int len);
+#else
+int utf8_naive(const unsigned char *data, int len);
+static inline int utf8_range2(const unsigned char *data, int len) {
```
fowles, line 6: `static inline` in a header file feels like an invitation for ODR violations if anyone ever takes the address of it

upb/decode_internal.h
```diff
@@ -70,24 +71,25 @@ const char *fastdecode_err(upb_decstate *d);
...
-    }
-    i += offset;
+bool decode_verifyutf8_inl(const char *ptr, int len) {
+  const char *end = ptr + len;
+
+  // Check 8 bytes at a time for any non-ASCII char.
+  while (end - ptr >= 8) {
+    uint64_t data;
+    memcpy(&data, ptr, 8);
+    if (data & 0x8080808080808080) break;
```
fowles, line 81: at this point you know it is not true and don't need to go into the second loop simply do

`return utf_range_2(ptr, end - potr) == 0`

upb/decode_internal.h
```diff
@@ -70,24 +71,25 @@ const char *fastdecode_err(upb_decstate *d);
...
+    ptr += 8;
   }
-  return i == len;
+
+  // Check one byte at a time for non-ASCII.
+  while (ptr < end) {
+    if (*ptr & 0x80) break;
```
fowles, line 87: don't bother with the break, early return on `return utf_range_2(ptr, end - ptr) == 0`.

upb/decode_internal.h
```diff
@@ -70,24 +71,25 @@ const char *fastdecode_err(upb_decstate *d);
...
+    ptr++;
+  }
+
+  return ptr == end ? true
+                    : utf8_range2((const unsigned char *)ptr, end - ptr) == 0;
```
fowles, line 92: and return true here

upb/decode_internal.h
haberman: Ditto.

third_party/utf8_range/utf8_range.h
haberman: This is a good resource that goes through C's inline rules and possible approaches. upb is using strategy (1): http://www.greenend.org.uk/rjk/tech/inline.html

fowles: From that link to parts come together for my concern:

- "(C++ is stricter: a function which is inline anywhere must be inline everywhere and must be defined identically in all the translation units that use it.)"
- In the strategy 1 section: "if you take the address of the function in two translation units then the result won't compare equal"

These two facts combine to make `static inline` ODR bait for folks taking the address of them.  I don't have a better answer given C's lack of stripping but that is the crux of my concern.

PR #430, opened 2021-10-12
cmake & conan

AnticliMaxtic:
I have reworked the CMake support for this project, and added a conanfile.py and test_package folder for [Conan](https://docs.conan.io) support. 

- CMake
  - created cmake/Modules/ProtocGenUpb.cmake so there is a CMake function that can be used to call the upb protoc plugin and generate upb files 
  - cmake/InstallHelpers.cmake, cmake/upbConfig.cmake.in, cmake/cmake_uninstall.cmake.in to generate install and uninstall targets  for calling `cmake --install` and `cmake --build . --target uninstall` or `ninja install`/`ninja uninstall` etc

- Conan 
  - conanfile.py written with basic instructions for getting dependencies and building
  - test_package for testing the "consumability" of the conan/cmake build

I am not familiar enough with Bazel to update the Bazel scripts that are used to generate the cmake files (could probably generate conan files as well), I can take a stab at it but would appreciate any help with that


cmake/CMakeLists.txt
```diff
@@ -3,33 +3,44 @@
...
 cmake_policy(SET CMP0048 NEW)
 
-project(upb)
+project(
+  upb
+  VERSION 0.0.0
+  LANGUAGES C CXX
+)
 set(CMAKE_C_STANDARD 99)
+set(CMAKE_CXX_STANDARD 17)
```
haberman, line 20: I think upb is targeting C++11 for the moment.

cmake/CMakeLists.txt
```diff
@@ -61,72 +68,134 @@ endif()
...
-  ../upb/table_internal.h
-  ../upb/upb.c
-  ../upb/upb_internal.h
-  ../upb/decode.h
-  ../upb/encode.h
-  ../upb/msg.h
-  ../upb/upb.h
-  ../upb/upb.hpp)
-target_link_libraries(upb
+list(APPEND UPB_INTERFACES port)
```
haberman, line 75: port is marked as an `INTERFACE` library because it has no sources, but it is not meant to be exported. Users should not be including `port_def.inc` or `port_undef.inc` directly.

To decide whether an interface should be exported, we should look at the visibility on the `cc_library()` rule in `make_cmakelists.py`.

cmake/Modules/ProtocGenUpb.cmake
```diff
@@ -0,0 +1,389 @@
...
+# The following variables can be set and are optional:
+#
+# PROTOBUF_SRC_ROOT_FOLDER - When compiling with MSVC, if this cache variable is
+#                            set the protobuf-default VS project build locations
+#                            (vsprojects/Debug & vsprojects/Release) will be
+#                            searched for libraries and binaries.
+#
+# PROTOC_PATH              - Path to protoc executable
+# UPB_PLUGIN_PATH          - Path to protoc-gen-upb executable
+# UPBDEF_PLUGIN_PATH       - Path to protoc-gen-upbdef executable
```
haberman, line 13: This should be `protoc-gen-upbdefs`, and maybe `UPBDEFS_PLUGIN_PATH`?

cmake/Modules/ProtocGenUpb.cmake
```diff
@@ -0,0 +1,389 @@
...
+#
+# Defines the following variables:
+#
+# The following cache variables are also available to set or use:
+# PROTOBUF_PROTOC_EXECUTABLE - The protoc compiler (full path if not found in ${PATH})
+# UPB_GENERATOR     - The upb plugin executable (full path if not found in ${PATH})
+# UPBDEFS_GENERATOR - The upbdefs plugin executable (full path if not found in ${PATH})
+#
+# =============================================================================
+# Copyright 2009 Kitware, Inc. Copyright 2009-2011 Philip Lowman
```
haberman, line 33: Google policy requires that all code that is not Copyright Google must go in the `third_party` directory.

cmake/Modules/ProtocGenUpb.cmake
```diff
@@ -0,0 +1,389 @@
...
+    return()
+  endif()
+
+  # Construct include list
+  set(_upb_include_set)
+  foreach(_dir ${UPB_INCLUDE_DIRS})
+    list(APPEND _upb_include_set -I${_dir})
+  endforeach()
+
+  if(UPB_GENERATE_X_APPEND_PATH)
```
haberman, line 199: This behavior is harmful; protos should always be referenced by their canonical filename. eg. `google/protobuf/descriptor.proto`, not just `descriptor.proto`.

In particular, only one file with a given name can be loaded at a time. If we have `foo/bar/baz.proto` and automatically add `-Ifoo/bar`, then `baz.proto` will conflict with any other file named `xyz/baz.proto`.

We should not implicitly add `-I` paths.

cmake/Modules/ProtocGenUpb.cmake
```diff
@@ -0,0 +1,389 @@
...
```
haberman, line 321: Ditto here, please don't add `-I` for every subdirectory.

haberman:
I'm going to close this for now, please feel free to re-open when you have addressed the comments above.

PR #432, opened 2021-10-20
Added a standalone required field check

haberman:
This is a follow-up to https://github.com/protocolbuffers/upb/pull/431. While the previous PR added required field checking in the parser, this adds a standalone pass to check for required fields. Compared with the parser check, this provides two important features:

1. It is perfectly accurate, containing neither false positives nor false negatives.
2. It can report the specific field path of the missing required field.

### DIFF
--- a/upb/util/required_fields.c
+++ b/upb/util/required_fields.c
@@ -27,6 +27,7 @@
 
 #include "upb/util/required_fields.h"
 
+#include <inttypes.h>
 #include <setjmp.h>
 #include <stdarg.h>
 #include <stdio.h>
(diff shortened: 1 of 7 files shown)


upb/util/required_fields.c
```diff
@@ -0,0 +1,294 @@
...
+  va_list args;
+
+  va_start(args, fmt);
+  n = vsnprintf(a->ptr, have, fmt, args);
+  va_end(args);
+
+  if (UPB_LIKELY(have > n)) {
+    a->ptr += n;
+  } else {
+    a->ptr = UPB_PTRADD(a->ptr, have);
```
fowles, line 64: Why is this line `PTRADD` and the one above (line 62) is just `+=`?

upb/util/required_fields.c
```diff
@@ -0,0 +1,294 @@
...
+static void upb_FieldPath_Printf(upb_PrintfAppender* a, const char* fmt, ...) {
+  size_t n;
+  size_t have = a->end - a->ptr;
+  va_list args;
+
+  va_start(args, fmt);
+  n = vsnprintf(a->ptr, have, fmt, args);
+  va_end(args);
+
+  if (UPB_LIKELY(have > n)) {
```
fowles, line 61: should this be >= ?

upb/util/required_fields.c
```diff
@@ -0,0 +1,294 @@
...
+  appender.end = UPB_PTRADD(buf, size);
+  appender.overflow = 0;
+  bool first = true;
+
+  while (ptr->field) {
+    const upb_fielddef *f = ptr->field;
+
+    upb_FieldPath_Printf(&appender, first ? "%s" : ".%s", upb_fielddef_name(f));
+    first = false;
+    ptr++;
```
fowles, line 129: why is this not `PTRADD(ptr, 1)` ?

upb/util/required_fields.c
```diff
@@ -0,0 +1,294 @@
...
+  appender.buf = buf;
+  appender.ptr = buf;
+  appender.end = UPB_PTRADD(buf, size);
+  appender.overflow = 0;
+  bool first = true;
+
+  while (ptr->field) {
+    const upb_fielddef *f = ptr->field;
+
+    upb_FieldPath_Printf(&appender, first ? "%s" : ".%s", upb_fielddef_name(f));
```
fowles, line 127: complex expressions in printf tends to defeat the checker.  this one is simple enough that you are fine, but you should like avoid them where convenient

upb/util/required_fields.c
```diff
@@ -0,0 +1,294 @@
...
+        memcpy(&ctx->out_fields.path[ctx->out_fields.size], ctx->stack.path,
+               ctx->stack.size * sizeof(*ctx->stack.path));
+        ctx->out_fields.size += ctx->stack.size;
+        ctx->out_fields.path[ctx->out_fields.size++] =
+            (upb_FieldPathEntry){.field = f};
+        ctx->out_fields.path[ctx->out_fields.size++] =
+            (upb_FieldPathEntry){.field = NULL};
+      }
+    }
+  }
```
fowles, line 225: can you extract this first loop to a well named helper function?

upb/util/required_fields.h
```diff
@@ -0,0 +1,90 @@
...
+//    int32_msg_map[5].bar       # Integer-keyed map
+//    string_msg_map["abc"]      # String-keyed map
+//    bool_msg_map[true]         # Bool-keyed map
+//
+// The input array `*path` must be NULL-terminated.  The pointer `*path` will be
+// updated to point to one past the terminating NULL pointer of the input array.
+//
+// The output buffer `buf` will always be NULL-terminated. If the output data
+// (including NULL terminator) exceeds `size`, the result will be truncated.
+size_t upb_FieldPath_ToText(upb_FieldPathEntry **path, char *buf, size_t size);
```
fowles, line 70: is the return value the size that would have been written had the buffer not filled?  if so please add that to the comment

upb/util/required_fields.c
haberman: Acknowledged.

upb/util/required_fields.c
haberman: `UPB_PTRADD()` is only needed when there is a possibility that the pointer is `NULL`. This pointer cannot be `NULL`; we just dereferenced it.

upb/util/required_fields.h
haberman: Done.

upb/util/required_fields.c
haberman: We can hit this branch if the user passed `upb_FieldPath_ToText(&foo, NULL, 0)`, whereas a 0 size cannot enter the previous branch.

fowles (approved):

PR #434, opened 2021-11-01
Added util library for converting defs back to descriptor protos

haberman:
These functions correspond to the `CopyTo()` family of functions in C++ descriptors: they convert a descriptor back to a descriptor proto.

We put these in a util library instead of `upb/def.c` because this functionality can easily be implemented on top of the core def APIs, and most users will not need it.

### DIFF
--- a/upb/util/def_to_proto.c
+++ b/upb/util/def_to_proto.c
@@ -13,11 +13,11 @@
  *       names of its contributors may be used to endorse or promote products
  *       derived from this software without specific prior written permission.
  *
- * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
- * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
- * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
- * DISCLAIMED. IN NO EVENT SHALL Google LLC BE LIABLE FOR ANY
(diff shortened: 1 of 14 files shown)


upb/util/def_to_proto.c
```diff
@@ -0,0 +1,507 @@
...
+google_protobuf_MethodDescriptorProto *upb_MethodDef_ToProto(
+    const upb_methoddef *m, upb_arena *a) {
+  toproto_ctx ctx = {a};
+  return UPB_SETJMP(ctx.err) ? NULL : methoddef_toproto(&ctx, m);
+}
+
+google_protobuf_ServiceDescriptorProto *upb_ServiceDef_ToProto(
+    const upb_servicedef *s, upb_arena *a) {
+  toproto_ctx ctx = {a};
+  return UPB_SETJMP(ctx.err) ? NULL : servicedef_toproto(&ctx, s);
```
fowles, line 506: macros in ternaries wig me out a bit....

upb/util/def_to_proto.h
```diff
@@ -0,0 +1,61 @@
...
+
+#include "upb/def.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// Functions for converting defs back to the equivalent descriptor proto.
+// Ultimately the goal is that a round-trip proto->def->proto is lossless.
+// Each function creates a
```
fowles, line 39: comment trails off a bit.  Also the other things in this file should be commented too

upb/util/def_to_proto.c
haberman: Done.

upb/util/def_to_proto.h
haberman: Done. I only made a single comment, since the one comment applies to all the functions that follow (they all follow the same form).

fowles (approved):

Issue #436, opened 2021-11-02
Fixed packed arrays encoding/decoding are not big endian compliant


stanhu:
The Lua binding tests are failing on a big endian system (e.g. 390x):

```
# cat /root/.cache/bazel/_bazel_root/6bf3da2484ef4bca9a0fbec97f98a05c/execroot/upb/bazel-out/s390x-fastbuild/testlogs/tests/bindings/lua/test_lua/test.log
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //tests/bindings/lua:test_lua
-----------------------------------------------------------------------------
    ..........F.....F.......F..

200382 Assertions checked.

error testing Lua: tests/bindings/lua/test_upb.lua:837: One or more errors in test suite
  1) Failure (testupb.test_string_double_map):
tests/bindings/lua/test_upb.lua:258: expected 102 but was 1711276032

  2) Failure (testupb.test_numeric_map):
tests/bindings/lua/test_upb.lua:647: error expected but no error occurred

  3) Failure (testupb.test_numeric_array):
tests/bindings/lua/test_upb.lua:588: error expected but no error occurred

Testsuite finished (24 passed, 3 failed, 0 errors).
```

I believe the lines in https://github.com/protocolbuffers/upb/blob/40dfcc28384ce2726cf8f126ec4056941aee2283/upb/encode.c#L299-L312 assume floats are in little endian format, so writing out the whole byte stream works fine.

I've validated this by writing the encoded binary out to a file:

```diff
diff --git a/tests/bindings/lua/test_upb.lua b/tests/bindings/lua/test_upb.lua
index 5f1aae8..ad00ee7 100644
--- a/tests/bindings/lua/test_upb.lua
+++ b/tests/bindings/lua/test_upb.lua
@@ -271,6 +271,10 @@ function test_string_double_map()
   check_msg(msg)

   local serialized_packed = upb.encode(msg)
+  local out = io.open("/build/fixed.bin", "wb")
+  out:write(serialized_packed)
+  out:close()
+
   local msg2 = upb.decode(upb_test.PackedTest, serialized_packed)
   local msg3 = upb.decode(upb_
```

### Bad

```
root@51e832e4e892:/build/upb# /build/protobuf/src/protoc -I$(pwd)/upb -I/build/protobuf -I/build/protobuf/src -I/build/upb/tests/bindings/lua --decode=upb_test.PackedTest test.proto < /build/fixed.bin
i32_packed: 100
[...]

PR #437, opened 2021-11-04
Fix big-endian encodings for fixed packed arrays

stanhu:
Previously the encoder would serialize the output buffer in one copy,
but this only works on little-endian systems. We now iterate through the
array and convert each value if necessary before serializing it out.

Part of https://github.com/protocolbuffers/upb/issues/436

### DIFF
--- a/upb/encode.c
+++ b/upb/encode.c
@@ -171,10 +171,23 @@ static void encode_fixedarray(upb_encstate *e, const upb_array *arr,
   size_t bytes = arr->len * elem_size;
   const char* data = _upb_array_constptr(arr);
   const char* ptr = data + bytes - elem_size;
-  if (tag) {
+
+  if (tag || !_upb_isle()) {
     while (true) {
-      encode_bytes(e, ptr, elem_size);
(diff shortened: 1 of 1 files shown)


stanhu:
@haberman Would you prefer this to be rewritten more like https://github.com/protocolbuffers/upb/pull/441?

haberman:
> @haberman Would you prefer this to be rewritten more like #441?

Yes please. On little-endian machines we should generate the same code we were already generating.

stanhu:
@haberman Ok, done.

stanhu:
@haberman I learned that you can easily spin up a s390x container with QEMU and Docker:

```
$ docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
$ docker run --rm -it s390x/ubuntu bash
```

More info here: https://github.com/multiarch/qemu-user-static

It's a little slow on my machine, and I'd imagine building `bazel` might take a while in an emulated environment. But this still seems useful for testing.

stanhu:
@haberman Would you mind reviewing this and #440?

PR #438, opened 2021-11-04
Fix big-endian decodings for fixed packed arrays

stanhu:
Previously the decoder would copy the input buffer as-is into the
deserialized array, but this only works on little-endian systems. We now
iterate through the buffer and convert each value if necessary.

Part of https://github.com/protocolbuffers/upb/issues/436

### DIFF
--- a/upb/decode.c
+++ b/upb/decode.c
@@ -434,23 +434,66 @@ static const char *decode_enum_toarray(upb_decstate *d, const char *ptr,
 }
 
 UPB_FORCEINLINE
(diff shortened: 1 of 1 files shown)


haberman:
I'd prefer to not change the code generated for little-endian machines. Could you try out https://github.com/protocolbuffers/upb/pull/441 and see if it works for you?

stanhu:
Yes, I just tested that, and it works. Closing this in favor of https://github.com/protocolbuffers/upb/pull/441.

Issue #439, opened 2021-11-04
LUA out-of-range checks may fail due to undefined compiler behavior


stanhu:
The following code does not work on a s390x system: https://github.com/protocolbuffers/upb/blob/dee1238cac20a633b970347d214995403717f6c5/upb/bindings/lua/upb.c#L196-L200

For example, let's use the `2^64` value: https://github.com/protocolbuffers/upb/blob/dee1238cac20a633b970347d214995403717f6c5/tests/bindings/lua/test_upb.lua#L429

This ends up converting to the `double` 18446744073709551616:

```c
#include <stdio.h>
#include <stdint.h>

int main() {
  double n = 18446744073709551616.0;
  uint64_t i = (uint64_t) n;

  printf("n = %f, i = %llu (0x%llx), (double) i = %f\n", n, i, i, (double) i);

    if ((double)i != n) {
    printf("out of range!\n");
  } else {
    printf("in range\n");
  }
}
```

On my macOS system, I see:

```
n = 18446744073709551616.000000, i = 0 (0x0), (double) i = 0.000000
out of range!
```

However, on an s390x platform, I see:

```
n = 18446744073709551616.000000, i = 18446744073709551615 (0xffffffffffffffff), (double) i = 18446744073709551616.000000
in range
```

As shown below, 18446744073709551616 is too large to be represented by a `uint64_t`:

```
test.c:7:24: error: integer literal is too large to be represented in any integer type
  uint64_t too_large = 18446744073709551616;
```

It looks like we get two different values depending on what system we're on. On macOS, 0 gets returned. On s390x, the value gets set to the highest unsigned 64-bit integer.

We might be hitting an integer overflow condition (https://frama-c.com/2013/10/09/Overflow-float-integer.html), and the compiler might be hitting undefined behavior as a result. For example, https://docs.microsoft.com/en-us/cpp/c-language/conversions-from-floating-point-types?view=msvc-160 says:

> Truncate at decimal point. If result is too large to be represented as unsigned long, result is undefined.

We might want to compare the max value for the specific type rather than rely on a cast.

stanhu:
As explained in https://stackoverflow.com/a/17822304, this naive patch doesn't work:

```diff
diff --git a/upb/bindings/lua/upb.c b/upb/bindings/lua/upb.c
index c9d98bb..c3adc32 100644
--- a/upb/bindings/lua/upb.c
+++ b/upb/bindings/lua/upb.c
@@ -52,6 +52,7 @@
[...]

PR #440, opened 2021-11-04
Fix LUA out-of-range checks to handle integer overflow

stanhu:
As discussed in https://github.com/protocolbuffers/upb/issues/439, the LUA binding tests were failing on some compilers (e.g. s390x) since conversion of out-of-range integers values (e.g. 2^64 or 18446744073709551616) to a double can lead to undefined behavior. On some compilers, this conversion can return 0, but it can also return the max value.

To make this check work across platforms, we do the following:

1. Break the double into its integral and fractional components.
2. If there any fractional components, return an error.
3. Compare the min and max values in an intelligent way. We can't simply
compare against UINT64_MAX since that value will get implicitly
converted to a double in an undefined way, and 18446744073709551616 >
UINT64_MAX can return false.  Instead, we divide the max value by 2, multiply it back, and add 1 to obtain the true out-of-bounds value as a double.

Closes https://github.com/protocolbuffers/upb/issues/439

### DIFF
--- a/upb/bindings/lua/upb.c
+++ b/upb/bindings/lua/upb.c
@@ -181,7 +181,7 @@ const char *lupb_checkstring(lua_State *L, int narg, size_t *len) {
 /* Unlike luaL_checkinteger, these do not implicitly convert from string or
  * round an existing double value.  We allow floating-point input, but only if
  * the actual value is integral. */
-#define INTCHECK(type, ctype, min, max_bits)                                   \
+#define INTCHECK(type, ctype, min, max)                                        \
   ctype lupb_check##type(lua_State *L, int narg) {                             \
     double n;                                                                  \
     if (lua_isinteger(L, narg)) {                                              \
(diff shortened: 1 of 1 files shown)


upb/bindings/lua/upb.c
```diff
@@ -207,10 +207,10 @@ const char *lupb_checkstring(lua_State *L, int narg, size_t *len) {
     lua_pushnumber(L, val);                                                    \
   }
 
-INTCHECK(int64,  int64_t)
-INTCHECK(int32,  int32_t)
-INTCHECK(uint64, uint64_t)
-INTCHECK(uint32, uint32_t)
+INTCHECK(int64,  int64_t, INT64_MIN, 63)
```
haberman, line 210: Can we use `max` (eg. `INT64_MAX`) instead of `maxbits`?

haberman: I would still prefer to use `INT64_MAX` instead of `maxbits`.

haberman:
@stanhu did you see my comment above?

stanhu:
@haberman Ok, this implementation now works on both x86 and s390x.

PR #441, opened 2021-11-05
Properly byte-swap fixed packed fields in the decoder

haberman:


stanhu:
This looks good to me!

Issue #442, opened 2021-11-05
DefToProto test seg faults on big endian systems


stanhu:
This seems to be coming from here: https://github.com/protocolbuffers/upb/blob/0cb7f72c02e899f58bc7e91b268b8077ca4b3601/upb/util/def_to_proto.c#L102-L104

I checked that right before it seg faults:

1. `upb_fielddef_fullname(f)` is `google.protobuf.FileOptions.optimize_for` 
2. `upb_fielddef_type(f)` is 5 (`UPB_DESCRIPTOR_TYPE_INT32`)
3. `d.int32_val` is 0
4.  `ev` is also 0

On my macOS build, `upb_enumdef_lookupnum` returns a non-zero pointer, whereas it returns 0 here. 

Digging deeper, it seems `upb_inttable_lookup` is returning 1 in the working case, 0 in the broken case. I have to wonder if the table implementation isn't hashing things with endian concerns.

`valgrind` output shows:

```
root@88265bec88fe:/build/upb# valgrind /root/.cache/bazel/_bazel_root/6bf3da2484ef4bca9a0fbec97f98a05c/execroot/upb/bazel-out/s390x-fastbuild/bin/upb/util/def_to_proto_test
==1540294== Memcheck, a memory error detector
==1540294== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1540294== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==1540294== Command: /root/.cache/bazel/_bazel_root/6bf3da2484ef4bca9a0fbec97f98a05c/execroot/upb/bazel-out/s390x-fastbuild/bin/upb/util/def_to_proto_test
==1540294==
Running main() from gmock_main.cc
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from DefToProto
[ RUN      ] DefToProto.Test
==1540294== Invalid read of size 8
==1540294==    at 0x53C3B9C: upb_enumvaldef_name (in /root/.cache/bazel/_bazel_root/6bf3da2484ef4bca9a0fbec97f98a05c/execroot/upb/bazel-out/s390x-fastbuild/bin/libreflection.so)
==1540294==    by 0x53ACA55: default_string (in /root/.cache/bazel/_bazel_root/6bf3da2484ef4bca9a0fbec97f98a05c/execroot/upb/bazel-out/s390x-fastbuild/bin/upb/util/libdef_to_proto.so)
[...]

PR #443, opened 2021-11-05
Fix encoding/decoding for def-to-proto on big-endian systems

stanhu:
In a big-endian system, the 64-bit value of 1 is represented as:

```
0x0 0x0 0x0 0x0 0x0 0x0 0x0 0x1
```

However, when `d.int32_val` is used, this truncates this and takes the
first four bytes:

```
0x0 0x0 0x0 0x0
```

As a result, we lose the value of 1 from this truncation and the value
beocmes 0. This doesn't happen in a little-endian system because the 1
is in the lowest memory address, so truncating the value to 32 bits
doesn't change anything.

Previously the DefToProto test was failing on a big-endian system
because this truncation caused the key to be incorrectly set to 0.

We now use the type-specific functions
(e.g. `upb_fielddef_defaultint32`) to do this conversion.

Closes https://github.com/protocolbuffers/upb/issues/442

### DIFF
--- a/upb/util/def_to_proto.c
+++ b/upb/util/def_to_proto.c
@@ -97,24 +97,24 @@ static upb_strview default_string(upb_ToProto_Context *ctx,
   upb_msgval d = upb_fielddef_default(f);
   switch (upb_fielddef_type(f)) {
     case UPB_TYPE_BOOL:
-      return strviewdup(ctx, d.bool_val ? "true" : "false");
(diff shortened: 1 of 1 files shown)


haberman:
I would prefer to fix the bug in `upb_fielddef_default()`, see: https://github.com/protocolbuffers/upb/issues/442#issuecomment-962259582

PR #444, opened 2021-11-05
[python] Added descriptor containers

haberman:
The Python reflection API exposes numerous sequences and maps that vend descriptor data. To avoid defining a new Python class for each one, we define a few abstract containers that take a "vtable" defining the specific kind of data that is returned.

The Python/C++ extension does a similar thing: https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/pyext/descriptor_containers.cc

Compared with the Python/C++ version, our design is a bit more conceptually clear, since we treat the three distinct container types separately instead of having a single unified vtable for all three.

### DIFF
--- a/python/descriptor_containers.c
+++ b/python/descriptor_containers.c
@@ -44,8 +44,13 @@ typedef struct {
   int index;             // Current iterator index.
 } PyUpb_DescriptorIterator;
 
-static void PyUpb_DescriptorIterator_Dealloc(PyUpb_DescriptorIterator *_self) {
-  PyUpb_DescriptorIterator *self = (PyUpb_DescriptorIterator*)_self;
+PyUpb_DescriptorIterator *PyUpb_DescriptorIterator_Self(PyObject *obj) {
+  assert(Py_TYPE(obj) == PyUpb_ModuleState_Get()->descriptor_iterator_type);
+  return (PyUpb_DescriptorIterator*)obj;
+}
+
+static void PyUpb_DescriptorIterator_Dealloc(PyObject *_self) {
+  PyUpb_DescriptorIterator *self = PyUpb_DescriptorIterator_Self(_self);
   Py_DECREF(self->parent_obj);
   PyUpb_Dealloc(self);
 }
@@ -65,7 +70,7 @@ PyObject *PyUpb_DescriptorIterator_New(const PyUpb_GenericSequence_Funcs *funcs,
 }
 
 PyObject* PyUpb_DescriptorIterator_IterNext(PyObject* _self) {
-  PyUpb_DescriptorIterator *self = (PyUpb_DescriptorIterator*)_self;
+  PyUpb_DescriptorIterator *self = PyUpb_DescriptorIterator_Self(_self);
   int size = self->funcs->get_elem_count(self->parent);
   if (self->index >= size) return NULL;
   const void *elem = self->funcs->index(self->parent, self->index);
@@ -99,6 +104,17 @@ typedef struct {
   PyObject *parent_obj;  // Python object that keeps parent alive, we own a ref.
 } PyUpb_GenericSequence;
 
+PyUpb_GenericSequence *PyUpb_GenericSequence_Self(PyObject *obj) {
+  assert(Py_TYPE(obj) == PyUpb_ModuleState_Get()->generic_sequence_type);
+  return (PyUpb_GenericSequence*)obj;
+}
+
+static void PyUpb_GenericSequence_Dealloc(PyObject *_self) {
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Self(_self);
+  Py_CLEAR(self->parent_obj);
+  PyUpb_Dealloc(self);
+}
+
 PyObject *PyUpb_GenericSequence_New(
     const PyUpb_GenericSequence_Funcs *funcs, const void *parent,
     PyObject *parent_obj) {
@@ -112,25 +128,14 @@ PyObject *PyUpb_GenericSequence_New(
   return &seq->ob_base;
 }
 
-PyUpb_GenericSequence *PyUpb_GenericSequence_Get(PyObject *obj) {
-  assert(Py_TYPE(obj) == PyUpb_ModuleState_Get()->generic_sequence_type);
-  return (PyUpb_GenericSequence*)obj;
-}
-
-static void PyUpb_GenericSequence_Dealloc(PyObject *_self) {
-  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
-  Py_CLEAR(self->parent_obj);
-  PyUpb_Dealloc(self);
-}
-
 static Py_ssize_t PyUpb_GenericSequence_Length(PyObject* _self) {
-  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Self(_self);
   return self->funcs->get_elem_count(self->parent);
 }
 
 static PyObject *PyUpb_GenericSequence_GetItem(PyObject *_self,
                                                Py_ssize_t index) {
-  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Self(_self);
   Py_ssize_t size = self->funcs->get_elem_count(self->parent);
   if (index < 0 || index >= size) {
     PyErr_Format(PyExc_IndexError, "list index (%zd) out of range", index);
@@ -162,9 +167,13 @@ static int PyUpb_GenericSequence_IsEqual(PyUpb_GenericSequence *self,
     PyObject *item1 = PyUpb_GenericSequence_GetItem((PyObject*)self, i);
     if (!item1) return -1;
     PyObject* item2 = PyList_GetItem(other, i);
-    if (!item2) return -1;
+    if (!item2) {
+      Py_DECREF(item1);
+      return -1;
+    }
     int cmp = PyObject_RichCompareBool(item1, item2, Py_EQ);
     Py_DECREF(item1);
+    Py_DECREF(item2);
     if (cmp != 1) return cmp;
   }
   // All items were found and equal
@@ -173,19 +182,19 @@ static int PyUpb_GenericSequence_IsEqual(PyUpb_GenericSequence *self,
 
 static PyObject *PyUpb_GenericSequence_RichCompare(PyObject *_self,
                                                    PyObject *other, int opid) {
-  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Self(_self);
   if (opid != Py_EQ && opid != Py_NE) {
-    Py_INCREF(Py_NotImplemented);
-    return Py_NotImplemented;
+    Py_RETURN_NOTIMPLEMENTED;
   }
-  bool ret = (opid == Py_EQ) == PyUpb_GenericSequence_IsEqual(self, other);
+  bool ret = PyUpb_GenericSequence_IsEqual(self, other);
+  if (opid == Py_NE) ret = !ret;
   return PyBool_FromLong(ret);
 }
 
-// Linear search.  Could optimize this in some, cases (defs that have index),
-// not not all (FileDescriptor.dependencies).
+// Linear search.  Could optimize this in some cases (defs that have index),
+// but not all (FileDescriptor.dependencies).
 static int PyUpb_GenericSequence_Find(PyObject *_self, PyObject *item) {
-  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Self(_self);
   const void *item_ptr = PyUpb_AnyDescriptor_GetDef(item);
   int count = self->funcs->get_elem_count(self->parent);
   for (int i = 0; i < count; i++) {
@@ -208,13 +217,17 @@ static PyObject* PyUpb_GenericSequence_Index(PyObject* self, PyObject* item) {
 
 // Implements list.count(): number of occurrences of the item in the sequence.
 // An item can only appear once in a sequence. If it exists, return 1.
-static PyObject *PyUpb_GenericSequence_Count(PyObject *self, PyObject *item) {
-  int position = PyUpb_GenericSequence_Find(self, item);
-  if (position < 0) {
-    return PyLong_FromLong(0);
-  } else {
-    return PyLong_FromLong(1);
+static PyObject *PyUpb_GenericSequence_Count(PyObject *_self, PyObject *item) {
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Self(_self);
+  const void *item_ptr = PyUpb_AnyDescriptor_GetDef(item);
+  int n = self->funcs->get_elem_count(self->parent);
+  int count = 0;
+  for (int i = 0; i < n; i++) {
+    if (self->funcs->index(self->parent, i) == item_ptr) {
+      count++;
+    }
   }
+  return PyLong_FromLong(count);
 }
 
 static PyObject *PyUpb_GenericSequence_Append(PyObject *self, PyObject *args) {
@@ -263,6 +276,17 @@ typedef struct {
   PyObject *parent_obj;  // Python object that keeps parent alive, we own a ref.
 } PyUpb_ByNameMap;
 
+PyUpb_ByNameMap *PyUpb_ByNameMap_Self(PyObject *obj) {
+  assert(Py_TYPE(obj) == PyUpb_ModuleState_Get()->by_name_map_type);
+  return (PyUpb_ByNameMap*)obj;
+}
+
+static void PyUpb_ByNameMap_Dealloc(PyObject *_self) {
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
+  Py_DECREF(self->parent_obj);
+  PyUpb_Dealloc(self);
+}
+
 PyObject *PyUpb_ByNameMap_New(const PyUpb_ByNameMap_Funcs *funcs,
                               const void *parent, PyObject *parent_obj) {
   PyUpb_ModuleState *s = PyUpb_ModuleState_Get();
@@ -274,19 +298,13 @@ PyObject *PyUpb_ByNameMap_New(const PyUpb_ByNameMap_Funcs *funcs,
   return &map->ob_base;
 }
 
-static void PyUpb_ByNameMap_Dealloc(PyObject *_self) {
-  PyUpb_ByNameMap *self = (void*)_self;
-  Py_DECREF(self->parent_obj);
-  PyUpb_Dealloc(self);
-}
-
 static Py_ssize_t PyUpb_ByNameMap_Length(PyObject* _self) {
-  PyUpb_ByNameMap *self = (void*)_self;
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
   return self->funcs->base.get_elem_count(self->parent);
 }
 
 static PyObject *PyUpb_ByNameMap_Subscript(PyObject *_self, PyObject *key) {
-  PyUpb_ByNameMap *self = (void*)_self;
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
   const char *name = PyUpb_GetStrData(key);
   const void *elem = name ? self->funcs->lookup(self->parent, name) : NULL;
 
@@ -306,14 +324,14 @@ static int PyUpb_ByNameMap_AssignSubscript(PyObject *self, PyObject *key,
 }
 
 static int PyUpb_ByNameMap_Contains(PyObject *_self, PyObject *key) {
-  PyUpb_ByNameMap *self = (void*)_self;
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
   const char *name = PyUpb_GetStrData(key);
   const void *elem = name ? self->funcs->lookup(self->parent, name) : NULL;
   return elem ? 1 : 0;
 }
 
 static PyObject *PyUpb_ByNameMap_Get(PyObject *_self, PyObject *args) {
-  PyUpb_ByNameMap *self = (void*)_self;
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
   PyObject* key;
   PyObject* default_value = Py_None;
   if (!PyArg_UnpackTuple(args, "get", 1, 2, &key, &default_value)) {
@@ -332,13 +350,13 @@ static PyObject *PyUpb_ByNameMap_Get(PyObject *_self, PyObject *args) {
 }
 
 static PyObject *PyUpb_ByNameMap_GetIter(PyObject *_self) {
-  PyUpb_ByNameMap *self = (PyUpb_ByNameMap *)_self;
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
   return PyUpb_DescriptorIterator_New(&self->funcs->base, self->parent,
                                       self->parent_obj);
 }
 
 static PyObject *PyUpb_ByNameMap_Keys(PyObject *_self, PyObject *args) {
-  PyUpb_ByNameMap *self = (PyUpb_ByNameMap *)_self;
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
   int n = self->funcs->base.get_elem_count(self->parent);
   PyObject *ret = PyList_New(n);
   if (!ret) return NULL;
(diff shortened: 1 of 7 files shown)


python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    PyObject *item = PyTuple_New(2);
+    if (!item) return NULL;
+    const void *elem = self->funcs->base.index(self->parent, i);
+    if (!elem) return NULL;
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!py_elem) return NULL;
```
fowles, line 379: pretty sure you are leaking `item` and `ret` on the early returns here

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+
+static PyObject *PyUpb_ByNameMap_Values(PyObject *_self, PyObject *args) {
+  PyUpb_ByNameMap *self = (PyUpb_ByNameMap *)_self;
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!elem) return NULL;
```
fowles, line 362: leaking `ret` here

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+}
+
+static PyObject *PyUpb_ByNameMap_RichCompare(PyObject *_self, PyObject *other,
+                                               int opid) {
+  PyUpb_ByNameMap *self = (void*)_self;
+  if (opid != Py_EQ && opid != Py_NE) {
+    Py_INCREF(Py_NotImplemented);
+    return Py_NotImplemented;
+  }
+  bool ret = (opid == Py_EQ) == PyUpb_ByNameMap_IsEqual(self, other);
```
fowles, line 414: same comment about formatting

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  }
+
+  const void *elem;
+  long num = PyLong_AsLong(key);
+  if (num == -1 && PyErr_Occurred()) {
+    elem = NULL;
+    PyErr_Clear();
+  } else {
+    elem = self->funcs->lookup(self->parent, num);
+  }
```
fowles, line 519: this long lookup dance is repeated verbatim above.  Seems like you should have a well named helper

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+
+static PyObject *PyUpb_ByNumberMap_Keys(PyObject *_self, PyObject *args) {
+  PyUpb_ByNumberMap *self = (PyUpb_ByNumberMap *)_self;
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *key = PyLong_FromLong(self->funcs->get_elem_num(elem));
+    if (!key) return NULL;
```
fowles, line 543: leak `ret`

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+static PyObject *PyUpb_ByNumberMap_Values(PyObject *_self, PyObject *args) {
+  PyUpb_ByNumberMap *self = (PyUpb_ByNumberMap *)_self;
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    if (!elem) return NULL;
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!py_elem) return NULL;
```
fowles, line 558: more leaks

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    PyObject *item = PyTuple_New(2);
+    if (!item) return NULL;
+    const void *elem = self->funcs->base.index(self->parent, i);
+    if (!elem) return NULL;
+    int number = self->funcs->get_elem_num(elem);
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!py_elem) return NULL;
```
fowles, line 576: more leaks

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  PyUpb_ByNumberMap *self = (PyUpb_ByNumberMap *)_self;
+  long num = PyLong_AsLong(key);
+  const void *elem;
+  if (num == -1 && PyErr_Occurred()) {
+    elem = NULL;
+    PyErr_Clear();
+  } else {
+    elem = self->funcs->lookup(self->parent, num);
+  }
+  return elem ? 1 : 0;
```
fowles, line 594: here is that pattern a third time

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+}
+
+static PyObject *PyUpb_ByNumberMap_RichCompare(PyObject *_self, PyObject *other,
+                                               int opid) {
+  PyUpb_ByNumberMap *self = (void*)_self;
+  if (opid != Py_EQ && opid != Py_NE) {
+    Py_INCREF(Py_NotImplemented);
+    return Py_NotImplemented;
+  }
+  bool ret = (opid == Py_EQ) == PyUpb_ByNumberMap_IsEqual(self, other);
```
fowles, line 623: same comment

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  Py_DECREF(self->parent_obj);
+  PyUpb_Dealloc(self);
+}
+
+PyObject *PyUpb_DescriptorIterator_New(const PyUpb_GenericSequence_Funcs *funcs,
+                                       const void *parent,
+                                       PyObject *parent_obj) {
+  PyUpb_ModuleState *s = PyUpb_ModuleState_Get();
+  PyUpb_DescriptorIterator *iter =
+      (void *)PyType_GenericAlloc(s->descriptor_iterator_type, 0);
```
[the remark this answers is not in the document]
haberman: The `s->descriptor_iterator_type` is a pointer to a type struct that contains the size (among other things).

The second argument `0` is `nitems`, intended for objects that are followed by the equivalent of a flexible array member. We don't use this feature, so the second argument is always zero.

https://docs.python.org/3/c-api/type.html#c.PyType_GenericAlloc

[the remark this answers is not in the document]
haberman: This is a common pattern in the Python C API, so I don't think it makes sense to comment every time. If you prefer, I could create a shared function in one of our .h files that contains a comment (and casts to `void*`).

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+      (PyUpb_GenericSequence *)PyType_GenericAlloc(s->generic_sequence_type, 0);
+  seq->funcs = funcs;
+  seq->parent = parent;
+  seq->parent_obj = parent_obj;
+  Py_INCREF(parent_obj);
+  return &seq->ob_base;
+}
+
+PyUpb_GenericSequence *PyUpb_GenericSequence_Get(PyObject *obj) {
+  assert(Py_TYPE(obj) == PyUpb_ModuleState_Get()->generic_sequence_type);
```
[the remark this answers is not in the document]
haberman: Done.

python/descriptor_containers.c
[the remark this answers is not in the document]
fowles: don't bother with the helper, that will be nearly mirroring the docs for the python api

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+static Py_ssize_t PyUpb_GenericSequence_Length(PyObject* _self) {
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
+  return self->funcs->get_elem_count(self->parent);
+}
+
+static PyObject *PyUpb_GenericSequence_GetItem(PyObject *_self,
+                                               Py_ssize_t index) {
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
+  Py_ssize_t size = self->funcs->get_elem_count(self->parent);
+  if (index < 0 || index >= size) {
```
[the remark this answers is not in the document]
haberman: In this case I'm not sure I agree.

As currently written, I read this as "if index is too small, or if it is too big." Index is consistently on the left because it's the variable whose validity is in question.

With your rewrite, it looks like we are first testing the validity of index, and then testing the validity of size against index. But size is known-good. For me personally your rewritten version is harder to visually scan.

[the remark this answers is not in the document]
fowles: I read it as "index is outside the range [0, size)" but if you feel otherwise feel free to leave as is.

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  // Check the identity of C++ pointers.
+  if (PyObject_TypeCheck(other, Py_TYPE(self))) {
+    PyUpb_GenericSequence *other_seq = (void *)other;
+    return self->parent == other_seq->parent && self->funcs == other_seq->funcs;
+  }
+
+  if (!PyList_Check(other)) return 0;
+
+  // return list(self) == other
+  int n = PyUpb_GenericSequence_Length((PyObject*)self);
```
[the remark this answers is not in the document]
haberman: Python lists are not restricted, but `PyUpb_GenericSequence` objects are (everything in the vtable uses `int` for index, following the conventions of `upb/def.h`.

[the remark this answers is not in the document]
fowles: but other might not be a upb list and so you might be truncating here

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  if (n != PyList_Size(other)) {
+    return false;
+  }
+  for (int i = 0; i < n; i++) {
+    PyObject *item1 = PyUpb_GenericSequence_GetItem((PyObject*)self, i);
+    if (!item1) return -1;
+    PyObject* item2 = PyList_GetItem(other, i);
+    if (!item2) return -1;
+    int cmp = PyObject_RichCompareBool(item1, item2, Py_EQ);
+    Py_DECREF(item1);
```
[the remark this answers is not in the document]
haberman: Done. (also freed `item1` in the failure path).

python/descriptor_containers.c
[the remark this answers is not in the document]
fowles: oh, wait, you are avoiding capturing other as an int.  This is a bit more subtle than desired though, please add a comment (or use size_t since you know it won't be negative)

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  }
+  // All items were found and equal
+  return 1;
+}
+
+static PyObject *PyUpb_GenericSequence_RichCompare(PyObject *_self,
+                                                   PyObject *other, int opid) {
+  PyUpb_GenericSequence *self = PyUpb_GenericSequence_Get(_self);
+  if (opid != Py_EQ && opid != Py_NE) {
+    Py_INCREF(Py_NotImplemented);
```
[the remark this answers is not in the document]
haberman: What does it mean for one list to be greater than or less than another list?

We could make it compare lengths. I'm not totally opposed to that. For now I was just going with the existing interface: https://github.com/protocolbuffers/protobuf/blob/3a4d9316aa9e3f0afec58e83ed744b0be4d337fa/python/google/protobuf/pyext/descriptor_containers.cc#L417-L420

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+    return Py_NotImplemented;
+  }
+  bool ret = (opid == Py_EQ) == PyUpb_GenericSequence_IsEqual(self, other);
```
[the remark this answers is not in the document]
haberman: Done.

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  return PyBool_FromLong(ret);
+}
+
+// Linear search.  Could optimize this in some, cases (defs that have index),
+// not not all (FileDescriptor.dependencies).
```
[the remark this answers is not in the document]
haberman: Done.

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+    PyErr_SetNone(PyExc_ValueError);
+    return NULL;
+  } else {
+    return PyLong_FromLong(position);
+  }
+}
+
+// Implements list.count(): number of occurrences of the item in the sequence.
+// An item can only appear once in a sequence. If it exists, return 1.
+static PyObject *PyUpb_GenericSequence_Count(PyObject *self, PyObject *item) {
```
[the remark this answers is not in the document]
haberman: Good question. The previous version of the code made this assumption, and it's *probably* true for all def sequences, but no need to rely on this. I rewrote to a more generic algorithm

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+}
+
+static void PyUpb_ByNameMap_Dealloc(PyObject *_self) {
+  PyUpb_ByNameMap *self = (void*)_self;
+  Py_DECREF(self->parent_obj);
+  PyUpb_Dealloc(self);
+}
+
+static Py_ssize_t PyUpb_ByNameMap_Length(PyObject* _self) {
+  PyUpb_ByNameMap *self = (void*)_self;
```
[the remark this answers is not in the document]
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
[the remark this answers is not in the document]
fowles: sorry, what I meant to say is "it seems odd that you have to incref on NotImplements

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
```diff
@@ -0,0 +1,670 @@
...
+  }
+
+  if (!PyList_Check(other)) return 0;
+
+  // return list(self) == other
+  int n = PyUpb_GenericSequence_Length((PyObject*)self);
+  if (n != PyList_Size(other)) {
+    return false;
+  }
+  for (int i = 0; i < n; i++) {
```
fowles, line 161: fun optimization trick:

Lists are very likely to defer in the first few elements or the last few elements.  It is rare for a list to only differ in the middle.  The last few element case can be thought of as comparing sorted lists.  As a result, comparing from the outside in is substantially faster than a linear scan in most real world settings.

fowles: ideally you would do it cacheline at a time, but that is way micro-optimizing

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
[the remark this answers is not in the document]
haberman: Ah, yep that is the Python C API. Every object, even global singleton objects like None or Py_NotImplemented.

I just saw that there is a macro for this. Changed to use that instead: https://docs.python.org/3/c-api/object.html#c.Py_RETURN_NOTIMPLEMENTED

python/descriptor_containers.c
haberman: Done.

haberman:
Addressed comments, PTAL.

python/descriptor_containers.c
```diff
@@ -208,13 +217,17 @@ static PyObject* PyUpb_GenericSequence_Index(PyObject* self, PyObject* item) {
 
 // Implements list.count(): number of occurrences of the item in the sequence.
 // An item can only appear once in a sequence. If it exists, return 1.
```
fowles, line 219: this comment is out of date

python/descriptor_containers.c
fowles: I actually think the first part is worth doing

python/descriptor_containers.c
```diff
@@ -0,0 +1,700 @@
...
+    return false;
+  }
+  for (int i = 0; i < n; i++) {
+    PyObject *item1 = PyUpb_GenericSequence_GetItem((PyObject*)self, i);
+    if (!item1) return -1;
+    PyObject* item2 = PyList_GetItem(other, i);
+    if (!item2) {
+      Py_DECREF(item1);
+      return -1;
+    }
```
fowles, line 173: I think I prefer the C idiom of:
```
Pyobject* item1 = NULL;
PyObject* item2 = NULL;
```
in the parent scope, then have the error cases do `goto error;` and close the function with
```
error:
  if (item1) Py_DECREF(item1);
  if (item2) Py_DECREF(item2);
  return -1;
```

python/descriptor_containers.c
```diff
@@ -0,0 +1,700 @@
...
+static PyObject *PyUpb_ByNameMap_Values(PyObject *_self, PyObject *args) {
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!elem) {
+      Py_DECREF(ret);
```
fowles, line 381: are you missing a decref on `py_elem` here? actually, it looks like you should be testing `py_elem` or should have this test one line higher

python/descriptor_containers.c
```diff
@@ -0,0 +1,700 @@
...
+    PyTuple_SetItem(item, 0,
+                    PyUnicode_FromString(self->funcs->get_elem_name(elem)));
+    PyTuple_SetItem(item, 1, py_elem);
+    PyList_SetItem(ret, i, item);
+  }
+  return ret;
+
+err2:
+  Py_DECREF(item);
+err1:
```
fowles, line 410: I would prefer to unify these labels to a single error label and let the compiler deal with the extra null checks

python/descriptor_containers.c
```diff
@@ -0,0 +1,700 @@
...
+static PyObject *PyUpb_ByNumberMap_Values(PyObject *_self, PyObject *args) {
+  PyUpb_ByNumberMap *self = PyUpb_ByNumberMap_Self(_self);
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!py_elem) {
+      Py_DECREF(ret);
```
fowles, line 587: very suspicious.  In the previous version you tested `elem` and not `py_elem`.

python/descriptor_containers.c
```diff
@@ -0,0 +1,700 @@
...
+    int number = self->funcs->get_elem_num(elem);
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!py_elem) goto err2;
+    PyTuple_SetItem(item, 0, PyLong_FromLong(number));
+    PyTuple_SetItem(item, 1, py_elem);
+    PyList_SetItem(ret, i, item);
+  }
+  return ret;
+
+err2:
```
fowles, line 614: same suggestion about a single error label

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Given the increase in complexity, this seems like premature optimization to me. There are several possible ways you could implement searching from the outsides in, which could have different performance characteristics and different edge cases to work through (empty list, etc). I'm not personally familiar with the empirical data that the middle elements are the least likely to be distinguishing. I'd rather punt on this for now.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Yes I meant to test py_elem.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: I realized that elem cannot be `NULL`, the functions we are calling assert that `i` is in bounds and will never return NULL. Only `py_elem` can by `NULL` when allocation fails.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
fowles: The relatively simple mechanism of

```
for (int i = 0; i < n/2; ++i) {
  compare(self[i], other[i])
  compare(self[n - i], other[n - i])
}
if (i % 2 == 1) {
  compare(self[i/2 + 1], other[i/2 + 1);
}
```

is a pretty easy and clean implementation.  I don't think it has serious complexity.

haberman:
PTAL.

python/descriptor_containers.c
```diff
@@ -391,24 +395,24 @@ static PyObject *PyUpb_ByNameMap_Items(PyObject *_self, PyObject *args) {
...
     PyList_SetItem(ret, i, item);
   }
   return ret;
 
-err2:
-  Py_DECREF(item);
-err1:
-  Py_DECREF(ret);
+error:
+  if (!py_elem) Py_DECREF(py_elem);
```
fowles, line 413: `if (py_elem)` not `if (!py_elem)`

haberman: I realized I could replace these with `Py_XDECREF()` for a more concise/foolproof NULL-check.

haberman:
PTAL.

python/descriptor_containers.c
fowles: of course that should be `n - 1 - i` cause offsets and all that.

fowles:
looks like the eq function for lists is the only thing left unresolved at this point.  I really think it is important to do because nearly sorted data is very common in practice and this changes eq from an average case O(n) to an average case O(1). 
 Everything else looks good.

python/descriptor_containers.c
```diff
@@ -0,0 +1,704 @@
...
+
+static PyObject *PyUpb_ByNameMap_Keys(PyObject *_self, PyObject *args) {
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *key = PyUnicode_FromString(self->funcs->get_elem_name(elem));
+    if (!key) return NULL;
```
fowles, line 370: leaking ret here

python/descriptor_containers.c
```diff
@@ -0,0 +1,704 @@
...
+  PyUpb_ByNameMap *self = PyUpb_ByNameMap_Self(_self);
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!py_elem) {
+      Py_DECREF(ret);
+      return NULL;
```
fowles, line 386: can we move this to the `goto error` idiom you uise elsewhere?

python/descriptor_containers.c
```diff
@@ -0,0 +1,704 @@
...
+static PyObject *PyUpb_ByNumberMap_Keys(PyObject *_self, PyObject *args) {
+  PyUpb_ByNumberMap *self = PyUpb_ByNumberMap_Self(_self);
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *key = PyLong_FromLong(self->funcs->get_elem_num(elem));
+    if (!key) {
+      Py_DECREF(ret);
```
fowles, line 574: more things that should go to the `goto error` idiom

python/descriptor_containers.c
```diff
@@ -0,0 +1,704 @@
...
+  PyUpb_ByNumberMap *self = PyUpb_ByNumberMap_Self(_self);
+  int n = self->funcs->base.get_elem_count(self->parent);
+  PyObject *ret = PyList_New(n);
+  if (!ret) return NULL;
+  for (int i = 0; i < n; i++) {
+    const void *elem = self->funcs->base.index(self->parent, i);
+    PyObject *py_elem = self->funcs->base.get_elem_wrapper(elem);
+    if (!py_elem) {
+      Py_DECREF(ret);
+      return NULL;
```
fowles, line 592: here too

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

python/descriptor_containers.c
haberman: Done.

fowles (approved):

PR #451, opened 2021-11-09
Conversion functions between Python and upb data types

haberman:
The functions in `python/convert.*` are state-less conversions between Python and upb data types.

These files also contain equality comparison for a few kinds of upb types.

### DIFF
--- a/python/convert.c
+++ b/python/convert.c
@@ -1,0 +1,327 @@
+/*
+ * Copyright (c) 2009-2021, Google LLC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of Google LLC nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL Google LLC BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "python/convert.h"
+
+#include "python/protobuf.h"
+#include "upb/reflection.h"
+
+PyObject* PyUpb_UpbToPy(upb_msgval val, const upb_fielddef *f, PyObject *arena) {
+  switch (upb_fielddef_type(f)) {
+    case UPB_TYPE_ENUM:
+    case UPB_TYPE_INT32:
+      return PyLong_FromLong(val.int32_val);
+    case UPB_TYPE_INT64:
+      return PyLong_FromLongLong(val.int64_val);
+    case UPB_TYPE_UINT32:
+      return PyLong_FromSize_t(val.uint32_val);
+    case UPB_TYPE_UINT64:
+      return PyLong_FromUnsignedLongLong(val.uint64_val);
+    case UPB_TYPE_FLOAT:
+      return PyFloat_FromDouble(val.float_val);
+    case UPB_TYPE_DOUBLE:
+      return PyFloat_FromDouble(val.double_val);
+    case UPB_TYPE_BOOL:
+      return PyBool_FromLong(val.bool_val);
+    case UPB_TYPE_BYTES:
+      return PyBytes_FromStringAndSize(val.str_val.data, val.str_val.size);
+    case UPB_TYPE_STRING:
+      return PyUnicode_DecodeUTF8(val.str_val.data, val.str_val.size, NULL);
+    case UPB_TYPE_MESSAGE:
+      PyErr_Format(PyExc_NotImplementedError,
+                   "Conversion of message types not yet implemented");
+    default:
+      PyErr_Format(PyExc_SystemError,
+                   "Getting a value from a field of unknown type %d",
+                   upb_fielddef_type(f));
+      return NULL;
+  }
+}
+
+static bool PyUpb_GetInt64(PyObject *obj, int64_t *val) {
+  // If the value is already a Python long, retrieves it.  Otherwise performs
+  // any automatic conversions to long (using __index__() or __int__()) that
+  // users expect.
+  *val = PyLong_AsLongLong(obj);
+  if (!PyErr_Occurred()) return true;
+  if (PyErr_ExceptionMatches(PyExc_OverflowError)) {
+    // Rewrite OverflowError -> ValueError.
+    // But don't rewrite other errors such as TypeError.
+    PyErr_Clear();
+    PyErr_Format(PyExc_ValueError, "Value out of range: %S", obj);
+  }
+  return false;
+}
+
+static bool PyUpb_GetUint64(PyObject *obj, uint64_t *val) {
+  // For uint64 Python does not offer any functions as convenient as
+  // PyLong_AsLongLong().  If the object is not already a "long" we must
+  // manually perform the automatic conversion (using __index__() or __int__())
+  // that users expect.
+  if (PyLong_Check(obj)) {
+    *val = PyLong_AsUnsignedLongLong(obj);
+  } else {
+    PyObject* casted = PyNumber_Long(obj);
+    if (!casted) return false;
+    *val = PyLong_AsUnsignedLongLong(casted);
+    Py_DECREF(casted);
+  }
+  if (!PyErr_Occurred()) return true;
+  PyErr_Clear();
+  PyErr_Format(PyExc_ValueError, "Value out of range: %S", obj);
+  return false;
+}
+
+// If `arena` is specified, copies the string data into the given arena.
+// Otherwise aliases the given data.
+static upb_msgval PyUpb_MaybeCopyString(const char *ptr, size_t size,
+                                        upb_arena *arena) {
+  upb_msgval ret;
+  ret.str_val.size = size;
+  if (arena) {
+    char *buf = upb_arena_malloc(arena, size);
+    memcpy(buf, ptr, size);
+    ret.str_val.data = buf;
+  } else {
+    ret.str_val.data = ptr;
+  }
+  return ret;
+}
+
+#define CHECK_RANGE(in, out, low, high) \
+  ((out) = (in), ((out) >= (low) && (out) <= (high)))
+
+bool PyUpb_PyToUpb(PyObject *obj, const upb_fielddef *f, upb_msgval *val,
+                   upb_arena *arena) {
+  switch (upb_fielddef_type(f)) {
+    case UPB_TYPE_ENUM:
+      if (PyUnicode_Check(obj)) {
+        Py_ssize_t size;
+        const char *name = PyUnicode_AsUTF8AndSize(obj, &size);
+        const upb_enumdef *e = upb_fielddef_enumsubdef(f);
+        const upb_enumvaldef *ev = upb_enumdef_lookupname(e, name, size);
+        if (!ev) {
+          PyErr_Format(PyExc_ValueError, "unknown enum label \"%s\"", name);
+          return false;
+        }
+        val->int32_val = upb_enumvaldef_number(ev);
+        return true;
+      } else {
+        if (!PyUpb_GetInt64(obj, &val->int64_val) ||
+            !CHECK_RANGE(val->int64_val, val->int32_val, INT32_MIN,
+                         INT32_MAX)) {
+          return false;
+        }
+        const upb_enumdef *e = upb_fielddef_enumsubdef(f);
+        if (upb_filedef_syntax(upb_enumdef_file(e)) == UPB_SYNTAX_PROTO2 &&
+            !upb_enumdef_checknum(e, val->int32_val)) {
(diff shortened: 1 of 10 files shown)


python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+    memcpy(buf, ptr, size);
+    ret.str_val.data = buf;
+  } else {
+    ret.str_val.data = ptr;
+  }
+  return ret;
+}
+
+#define CHECK_RANGE(in, out, low, high) \
+  ((out) = (in), ((out) >= (low) && (out) <= (high)))
```
fowles, line 116: I really don't like having this as a macro.  In particular it evaluates out 3 different times, which is very dangerous

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+        }
+        const upb_enumdef *e = upb_fielddef_enumsubdef(f);
+        if (upb_filedef_syntax(upb_enumdef_file(e)) == UPB_SYNTAX_PROTO2 &&
+            !upb_enumdef_checknum(e, val->int32_val)) {
+          PyErr_Format(PyExc_ValueError, "invalid enumerator %d",
+                       (int)val->int32_val);
+          return false;
+        }
+        return true;
+      }
```
fowles, line 147: can you split the longer cases into helper functions?

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+        const upb_enumvaldef *ev = upb_enumdef_lookupname(e, name, size);
+        if (!ev) {
+          PyErr_Format(PyExc_ValueError, "unknown enum label \"%s\"", name);
+          return false;
+        }
+        val->int32_val = upb_enumvaldef_number(ev);
+        return true;
+      } else {
+        if (!PyUpb_GetInt64(obj, &val->int64_val) ||
+            !CHECK_RANGE(val->int64_val, val->int32_val, INT32_MIN,
```
fowles, line 135: the in vs out of this macro is also not obvious

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+    default:
+      return false;
+  }
+}
+
+bool PyUpb_Map_IsEqual(const upb_map *map1, const upb_map *map2,
+                       const upb_fielddef *f) {
+  assert(upb_fielddef_ismap(f));
+  const upb_msgdef *entry_m = upb_fielddef_msgsubdef(f);
+  const upb_fielddef *val_f = upb_msgdef_field(entry_m, 1);
```
fowles, line 242: I think you should move this msgsubdef/msgdef_field below the fast path rejections based on pointer and size

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  return true;
+}
+
+static bool PyUpb_ArrayElem_IsEqual(const upb_array *arr1,
+                                    const upb_array *arr2, size_t i,
+                                    const upb_fielddef *f) {
+  upb_msgval val1 = upb_array_get(arr1, i);
+  upb_msgval val2 = upb_array_get(arr2, i);
```
fowles, line 266: do you need a bounds check here?

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+
+  if (arr1 == arr2) return true;
+  if (n1 != n2) return false;
+
+  // Half the length rounded up.  Important: the empty list rounds to 0.
+  size_t half = (n1 + 1) / 2;
+
+  // Search from the ends-in.  We expect differences to more quickly manifest
+  // at the ends than in the middle.  We may compare the most middle element
+  // twice.
```
fowles, line 284: I slightly prefer the formulation that does not compare the middle element twice since a full comparison on a proto message may be expensive

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+  if (msg1 == msg2) return true;
+
+  while (true) {
+    const upb_fielddef *f1, *f2;
+    upb_msgval val1, val2;
+    bool ok1 = msg1 && upb_msg_next(msg1, m, NULL, &f1, &val1, &iter1);
+    bool ok2 = msg2 && upb_msg_next(msg2, m, NULL, &f2, &val2, &iter2);
+    if (ok1 != ok2) return false;
+    if (!ok1) break;  // Both messages are at end.
+    if (f1 != f2) return false;
```
fowles, line 307: can you add a comment here saying "if the fields have different types" I spent a while puzzling over why the rest was useful is there was already pointer identity

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+      if (!PyUpb_ValueEq(val1, val2, f1)) return false;
+    }
+  }
+
+  size_t usize1, usize2;
+  const char *uf1 = upb_msg_getunknown(msg1, &usize1);
+  const char *uf2 = upb_msg_getunknown(msg2, &usize2);
+  // TODO(haberman): compare unknown fields one by one.
+  if (usize1 != usize2) return false;
+  if (memcmp(uf1, uf2, usize1)) return false;
```
fowles, line 324: pretty sure this memcmp is not valid.  You can have different binary unknown field reps that should compare equal

python/convert.h
```diff
@@ -0,0 +1,61 @@
...
+#include "protobuf.h"
+
+// Converts `val` to a Python object according to the type information in `f`.
+// Any newly-created objects that reference non-primitive data from `val` should
+// reference `arena` to keep the referent alive.  If the conversion cannot be
+// performed, returns NULL and sets a Python error.
+PyObject *PyUpb_UpbToPy(upb_msgval val, const upb_fielddef *f, PyObject *arena);
+
+// Converts `obj` to a upb_msgval `*val` according to the type information in
+// `f`. If `arena is provided, any string data will be copied into `arena`,
```
fowles, line 43: the first arena in this line is missing a closing back tick

python/convert.c
```diff
@@ -0,0 +1,327 @@
...
+                   "Getting a value from a field of unknown type %d",
+                   upb_fielddef_type(f));
+      return NULL;
+  }
+}
+
+static bool PyUpb_GetInt64(PyObject *obj, int64_t *val) {
+  // If the value is already a Python long, retrieves it.  Otherwise performs
+  // any automatic conversions to long (using __index__() or __int__()) that
+  // users expect.
```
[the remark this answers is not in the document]
haberman: The `PyLong_AsLongLong()` function does all of this internally. Updated the comment to make this more clear.

python/convert.c
haberman: Done.

python/convert.c
haberman: Done.

python/convert.c
haberman: Done.

python/convert.c
haberman: Done.

python/convert.c
haberman: This is only called from below, which statically verifies the bounds. I added an assert to clarify.

python/convert.c
haberman: Done (there is now a `upb/util/compare.c` that expands into the equivalent of an UnknownFieldSet to do a sorted, "canonical" representation before comparing).

python/convert.h
haberman: Done.

python/convert.c
haberman: Added a comment (my explanation for this is somewhat different than your suggestion).

haberman:
PTAL

python/convert.c
```diff
@@ -0,0 +1,344 @@
...
+      return PyFloat_FromDouble(val.double_val);
+    case UPB_TYPE_BOOL:
+      return PyBool_FromLong(val.bool_val);
+    case UPB_TYPE_BYTES:
+      return PyBytes_FromStringAndSize(val.str_val.data, val.str_val.size);
+    case UPB_TYPE_STRING:
+      return PyUnicode_DecodeUTF8(val.str_val.data, val.str_val.size, NULL);
+    case UPB_TYPE_MESSAGE:
+      PyErr_Format(PyExc_NotImplementedError,
+                   "Conversion of message types not yet implemented");
```
fowles, line 57: can you `return NULL` after this Err as well?

python/convert.c
haberman: Done.

python/convert.c
```diff
@@ -0,0 +1,344 @@
...
+      return false;
+    }
+    val->int32_val = upb_enumvaldef_number(ev);
+    return true;
+  } else {
+    if (!PyUpb_GetInt64(obj, &val->int64_val) ||
+        val->int64_val < INT32_MIN || val->int64_val > INT32_MAX) {
+      return false;
+    }
+    val->int32_val = val->int64_val;
```
fowles, line 133: I have a slight worry that users will be able to depend on the int32_val being set even in the error case.  Can we move it below the error check and keep a local here?

fowles: (same for the int64_val part of it

python/convert.c
```diff
@@ -0,0 +1,344 @@
...
+                   upb_arena *arena) {
+  switch (upb_fielddef_type(f)) {
+    case UPB_TYPE_ENUM:
+      return PyUpb_PyToUpbEnum(obj, upb_fielddef_enumsubdef(f), val);
+    case UPB_TYPE_INT32:
+      if (!PyUpb_GetInt64(obj, &val->int64_val) || val->int64_val < INT32_MIN ||
+          val->int64_val > INT32_MAX) {
+        return false;
+      }
+      val->int32_val = val->int64_val;
```
fowles, line 154: this replicates the error checking code of the enum above.  Maybe share something?

python/convert.c
```diff
@@ -0,0 +1,344 @@
...
+      return true;
+    case UPB_TYPE_INT64:
+      return PyUpb_GetInt64(obj, &val->int64_val);
+    case UPB_TYPE_UINT32:
+      if (!PyUpb_GetUint64(obj, &val->uint64_val) ||
```
fowles, line 159: more visible side effects on the error path

python/convert.c
```diff
@@ -0,0 +1,344 @@
...
+  }
+}
+
+bool PyUpb_Map_IsEqual(const upb_map *map1, const upb_map *map2,
+                       const upb_fielddef *f) {
+  assert(upb_fielddef_ismap(f));
+  size_t size1 = map1 ? upb_map_size(map1) : 0;
+  size_t size2 = map2 ? upb_map_size(map2) : 0;
+
+  if (map1 == map2) return true;
```
fowles, line 253: you can lift the pointer check above the size call

python/convert.c
```diff
@@ -0,0 +1,344 @@
...
+  return PyUpb_ValueEq(val1, val2, f);
+}
+
+bool PyUpb_Array_IsEqual(const upb_array *arr1, const upb_array *arr2,
+                         const upb_fielddef *f) {
+  assert(upb_fielddef_isseq(f) && !upb_fielddef_ismap(f));
+  size_t n1 = arr1 ? upb_array_size(arr1) : 0;
+  size_t n2 = arr2 ? upb_array_size(arr2) : 0;
+
+  if (arr1 == arr2) return true;
```
fowles, line 288: pointer check before size?

python/convert.c
```diff
@@ -0,0 +1,344 @@
...
+      }
+    } else {
+      if (!PyUpb_ValueEq(val1, val2, f1)) return false;
+    }
+  }
+
+  size_t usize1, usize2;
+  const char *uf1 = upb_msg_getunknown(msg1, &usize1);
+  const char *uf2 = upb_msg_getunknown(msg2, &usize2);
+  return upb_Message_UnknownFieldsAreEqual(uf1, usize1, uf2, usize2, 100);
```
fowles, line 343: 100 feels a bit magical.  Add a comment?

python/convert.h
```diff
@@ -0,0 +1,61 @@
...
+#ifndef PYUPB_CONVERT_H__
+#define PYUPB_CONVERT_H__
+
+#include "upb/def.h"
+#include "upb/reflection.h"
+
+#include "protobuf.h"
+
+// Converts `val` to a Python object according to the type information in `f`.
+// Any newly-created objects that reference non-primitive data from `val` should
```
fowles, line 37: should in comments is a little confusing.  Does it mean "the user must do this" or does it mean "the function must do this, but could be buggy"

upb/util/compare.c
```diff
@@ -0,0 +1,301 @@
...
+
+  *base = upb_arena_realloc(ctx->arena, *base, old * sizeof(**base),
+                            new * sizeof(**base));
+  if (!*base) UPB_LONGJMP(ctx->err, kUpb_UnknownCompareResult_OutOfMemory);
+
+  *ptr = *base + old;
+  *end = *base + new;
+}
+
+static const char *upb_UnknownFields_ParseVarint(const char *ptr,
```
fowles, line 79: I am greatly surprised you don't have a varint parser kicking around elsewhere in upb

upb/util/compare.h
```diff
@@ -0,0 +1,66 @@
...
+// and varints are made canonical.
+//
+// This operation should be considered best effort; the comparison is inherently
+// lossy without schema data:
+//
+//  1. We don't know whether delimited fields are sub-messages. Unknown
+//     sub-messages will therefore not have their fields sorted and varints
+//     canonicalized.
+//  2. We don't know about oneof/non-repeated fields, which should semantically
+//     discard every value except the last.
```
fowles, line 47: I think we want to take a slightly stronger stance here and even recommend users not call this code, because it is so best effort

upb/util/compare_test.cc
```diff
@@ -0,0 +1,255 @@
...
+  return ret;
+}
+
+TypeAndValue Group(UnknownFields nested) {
+  TypeAndValue ret{UnknownFieldType::kGroup};
+  ret.value.group = &nested;
+  return ret;
+}
+
+void f() {
```
fowles, line 107: what is the point of this function?

upb/util/compare_test.cc
```diff
@@ -0,0 +1,255 @@
...
+TEST(CompareTest, LongVarint) {
+  EXPECT_EQ(kUpb_UnknownCompareResult_Equal,
+            CompareUnknown({{1, LongVarint(123)}, {2, LongVarint(456)}},
+                           {{1, Varint(123)}, {2, Varint(456)}}));
+  EXPECT_EQ(kUpb_UnknownCompareResult_Equal,
+            CompareUnknown({{2, LongVarint(456)}, {1, LongVarint(123)}},
+                           {{1, Varint(123)}, {2, Varint(456)}}));
+}
+
+TEST(CompareTest, MaxDepth) {
```
fowles, line 248: we should add a fuzzer for this eventually, but not now.

python/convert.c
haberman: Done.

python/convert.c
haberman: Done.

python/convert.c
haberman: Done.

python/convert.c
haberman: Done.

python/convert.c
haberman: Done.

upb/util/compare_test.cc
haberman: Whoops, a test function left in. Deleted.

upb/util/compare_test.cc
haberman: Acknowledged.

upb/util/compare.c
haberman: I don't really want to support a public API for this in the core, since the function signature depends on the design of the parser and could change. I also would prefer not to have a util library depending on a private API. If there are two different util libraries that both need this, we could consider adding a util library for it, but it's so little code that it doesn't buy very much.

python/convert.h
haberman: Done.

python/convert.c
haberman: Done.

haberman:
PTAL.

fowles (approved):

PR #454, opened 2021-11-17
Check extension field numbers against extension ranges.

haberman:
This makes extension checking more strict in most cases.
However it also fixes a bug with MessageSet where we were being
too strict.  MessageSet allows larger extension numbers than
normal extensions do.

The change looks bigger than it is due to some refactoring of
`resolve_fielddef()` that breaks it apart into smaller functions.

### DIFF
--- a/upb/def.c
+++ b/upb/def.c
@@ -2178,7 +2178,7 @@ static void create_fielddef(
   const char *full_name;
   const char *json_name;
   const char *shortname;
-  uint32_t field_number;
+  int32_t field_number;
 
   f->file = ctx->file;  /* Must happen prior to symtab_add(). */
 
@@ -2241,7 +2241,7 @@ static void create_fielddef(
     upb_value v, field_v, json_v, existing_v;
     size_t json_size;
 
-    if (field_number == 0 || field_number > UPB_MAX_FIELDNUMBER) {
+    if (field_number <= 0 || field_number > UPB_MAX_FIELDNUMBER) {
       symtab_errf(ctx, "invalid field number (%u)", field_number);
     }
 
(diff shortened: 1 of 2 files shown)


upb/def.c
```diff
@@ -2245,6 +2241,10 @@ static void create_fielddef(
     upb_value v, field_v, json_v, existing_v;
     size_t json_size;
 
+    if (field_number == 0 || field_number > UPB_MAX_FIELDNUMBER) {
```
fowles, line 2244: I vaguelly prefer <= so I don't need to remember if this is signed or not

upb/def.c
```diff
@@ -2693,36 +2691,60 @@ static void resolve_fielddef(symtab_addctx *ctx, const char *prefix,
       // No resolution necessary.
       break;
   }
+}
+
+static void resolve_extension(
+    symtab_addctx *ctx, const char *prefix, upb_fielddef *f,
+    const google_protobuf_FieldDescriptorProto *field_proto) {
+  if (!google_protobuf_FieldDescriptorProto_has_extendee(field_proto)) {
+    symtab_errf(ctx, "extension for field '%s' had no extendee",
```
fowles, line 2700: I am guessing this contains a longjmp or something, but would lightly prefer a return so I don't have to worry about it

fowles (approved):

upb/def.c
haberman: Done (actually I changed it to `int32_t` since the field is int32 and we should in fact check for and reject negative values here.

upb/def.c
haberman: Unfortunately that would require a separate `return` statement. C with `-pedantic` doesn't allow `return f()` in a function returning `void`.

Another option is to make all the functions in this file return a `bool`, even though that return value will never actually used.

Another option is to drop `-pedantic`, though that could compromise the portability.

None of these options seem great. :/

fowles (approved):

PR #455, opened 2021-11-19
Upgrade Bazel workspace to use protobuf v3.19.1

stanhu:
Builds were not working because the mirror URL for v3.15.3 no longer
exists.


stanhu:
Closing since https://github.com/protocolbuffers/upb/pull/465 was merged.

PR #461, opened 2021-12-06
Message and MessageMeta classes for Python

haberman:
This is the core of the message implementation for Python. This contains both the message type itself (`CMessage`), as well as the message meta-type (`MessageMeta`) which provides special attributes on message classes.

There is a significant workaround for [a Python bug](https://bugs.python.org/issue37879) in `PyUpb_CMessage_Dealloc`. This was very painful to track down, but the fix is robust as long as we check the Python version and only apply the workaround to Python <3.8.

There is some cleverness in how we work around some limitations of the [limited API](https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface) (see `PyUpb_CPythonBits`), but these techniques should be robust. The limited API is very worthwhile, as it allows a single compiled extension to be ABI-compatible with all versions of Python that we support.


python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+}
+
+static PyObject* PyUpb_CMessage_RichCompare(PyObject* _self, PyObject* other,
+                                            int opid) {
+  PyUpb_CMessage* self = (PyUpb_CMessage*)_self;
+  if (opid != Py_EQ && opid != Py_NE) {
+    Py_INCREF(Py_NotImplemented);
+    return Py_NotImplemented;
+  }
+  bool ret = (opid == Py_EQ) == PyUpb_CMessage_IsEqual(self, other);
```
fowles, line 575: I think this structure is a bit more clever than my taste.  Can we do this with two statements?
```
bool ret = IsEqual(...);
if (opid == NE) ret = !ret;
return ret;
```

https://godbolt.org/z/xfKexdE9M

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+    // lead to a double-decref: https://bugs.python.org/issue37879
+    Py_DECREF(tp);
+  }
+}
+
+PyObject* PyUpb_CMessage_Get(upb_msg* u_msg, const upb_msgdef* m,
+                             PyObject* arena) {
+  PyObject* ret = PyUpb_ObjCache_Get(u_msg);
+
+  if (!ret) {
```
fowles, line 626:
```
if (ret) return ret;

...
return ret;
```

allows you to avoid some nesting

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  } else {
+    upb_msgval val;
+    if (PyUpb_CMessage_IsUnset(self)) {
+      // Unset message always returns default values.
+      val = upb_fielddef_default(field);
+    } else {
+      val = upb_msg_get(self->msg, field);
+    }
+    return PyUpb_UpbToPy(val, field, self->arena);
+  }
```
fowles, line 722: can you pull a few well named helpers out of this?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+
+  // Check base class attributes.
+  assert(!PyErr_Occurred());
+  PyObject* ret = PyObject_GenericGetAttr(_self, attr);
+
+  // Return value if found, swallow AttributeError if raised.
+  if (ret) {
+    return ret;
+  }
+  if (!PyErr_ExceptionMatches(PyExc_AttributeError)) {
```
fowles, line 780: I think this is a bit more clear as:

```
if (ret) return ret;
if (Matches(AttributeError)) {
  // swallow attribute errors because XXX
  PyErr_Clear();
  return PyUpb...
}

return NULL;
```

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  }
+
+  if (PyUpb_CMessage_IsUnset(self)) Py_RETURN_FALSE;
+
+  return PyBool_FromLong(field ? upb_msg_has(self->msg, field)
+                               : upb_msg_whichoneof(self->msg, oneof) != NULL);
+}
+
+static PyObject* PyUpb_CMessage_ListFields(PyObject* _self, PyObject* arg) {
+  PyObject* list = PyList_New(0);
```
fowles, line 828: Can you push this into the branches so we can give a size estimate to PyList to start with?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+                 "Parameter to MergeFrom() must be instance of same class: "
+                 "expected %S got %S.",
+                 Py_TYPE(self), Py_TYPE(arg));
+    return NULL;
+  }
+  // OPT: exit if src is empty.
+  PyObject* subargs = PyTuple_New(0);
+  PyObject* serialized = PyUpb_CMessage_SerializeToString(arg, subargs, NULL);
+  Py_DECREF(subargs);
+  if (!serialized) return NULL;
```
fowles, line 861: do you need to set an error for this?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+
+PyObject* PyUpb_CMessage_MergeFromString(PyObject* _self, PyObject* arg) {
+  PyUpb_CMessage* self = (PyUpb_CMessage*)_self;
+  char* buf;
+  Py_ssize_t size;
+  PyObject* bytes = NULL;
+
+  if (PyMemoryView_Check(arg)) {
+    bytes = PyBytes_FromObject(arg);
+    PyBytes_AsStringAndSize(bytes, &buf, &size);
```
fowles, line 889: why do you not have to check the return code of this function?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+    size_t size = 0;
+    size_t i = 0;
+    while (fields) {
+      upb_FieldPathEntry* field = fields;
+      size_t need = upb_FieldPath_ToText(&fields, buf, size);
+      if (need >= size) {
+        fields = field;
+        size = 16;
+        while (size <= need) size *= 2;
+        buf = realloc(buf, size);
```
fowles, line 1027: what about `std::min(size*2, need)`?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+        need = upb_FieldPath_ToText(&fields, buf, size);
+        assert(size > need);
```
fowles, line 1029: why strictly`>`?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+static PyObject* PyUpb_CMessage_FromString(PyObject* cls,
+                                           PyObject* serialized) {
+  PyObject* ret = PyObject_CallObject(cls, NULL);
+  if (ret == NULL) return NULL;
+
+  PyObject* length = PyUpb_CMessage_MergeFromString(ret, serialized);
+  if (length == NULL) {
+    Py_DECREF(ret);
+    return NULL;
+  }
```
fowles, line 1047: can we common out the error handling at the bottom?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  for (size_t i = 0; i < len; i++) {
+    if (filename[i] == '/') {
+      modname[i] = '.';
+    } else {
+      modname[i] = filename[i];
+    }
+  }
+  modname[len] = '\0';
+  PyObject* ret = PyUnicode_FromFormat("%s_pb2.%s", modname, msgname);
+  free(modname);
```
fowles, line 1297: you know max path length from filesystem constants I think... we could avoid allocating here

python/message.h
```diff
@@ -0,0 +1,93 @@
...
+
+#include <stdbool.h>
+
+#include "python/protobuf.h"
+#include "upb/reflection.h"
+
+// Removes the wrapper object for this field from the unset subobject cache.
+void PyUpb_CMessage_CacheDelete(PyObject* _self, const upb_fielddef* f);
+
+// Sets the field value for `f` to `subobj`, evicting the wrapper object from
```
fowles, line 39: here you use markdown.  Elsewhere you use doxygen.  I prefer markdown, but you should be consistent across the whole project.  (feel free to put this in the bucket of follow on global cleanups)

python/message.h
```diff
@@ -0,0 +1,93 @@
...
+void PyUpb_CMessage_SetConcreteSubobj(PyObject* _self, const upb_fielddef* f,
+                                      upb_msgval subobj);
+
+// Gets a Python wrapper object for message `u_msg` of type `m`, returning a
+// cached wrapper if one was previously created.  If a new object is created,
+// it will reference `arena`, which must own `u_msg`.
+PyObject* PyUpb_CMessage_Get(upb_msg* u_msg, const upb_msgdef* m,
+                             PyObject* arena);
+
+// Call to check whether the given Python object is a message.  If not, returns
```
fowles, line 51: Verifies that a python object is a message.  Sets a TypeError exception and returns false on failure.

python/protobuf.c
```diff
@@ -68,34 +68,79 @@ PyUpb_ModuleState *PyUpb_ModuleState_Get(void) {
...
+    PyObject *wkt_module =
+        PyImport_ImportModule("google.protobuf.internal.well_known_types");
+
+    if (wkt_module == NULL) {
+      return false;
+    }
+
+    state->wkt_bases = PyObject_GetAttrString(wkt_module, "WKTBASES");
+    PyObject *m = PyState_FindModule(&module_def);
+    // Make sure it is GC'd.
```
fowles, line 82: is or isn't?

python/protobuf.c
```diff
@@ -68,34 +68,79 @@ PyUpb_ModuleState *PyUpb_ModuleState_Get(void) {
...
+}
+
+void PyUpb_WeakMap_Add(PyUpb_WeakMap *map, const void *key, PyObject *py_obj) {
+  /*
+  uintptr_t k = (uintptr_t)key;
+  for (int i = 0; i < 64; i++) {
+    counts[i] += ((k & (1ULL << i)) != 0);
   }
+  total++;
+  */
```
fowles, line 124: commented out code

python/protobuf.c
```diff
@@ -104,6 +149,52 @@ PyObject *PyUpb_ObjCache_Get(const void *key) {
   }
 }
 
+bool PyUpb_WeakMap_Next(PyUpb_WeakMap *map, const void **key, PyObject **obj,
+                        intptr_t *iter) {
+  uintptr_t u_key;
+  upb_value val;
+  if (upb_inttable_next2(&map->table, &u_key, &val, iter)) {
```
fowles, line 156: I would invert this conditional and denest the happy path

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+      0,  // To be filled in by size of base     // tp_basicsize
+      0,  // tp_itemsize
+      Py_TPFLAGS_DEFAULT,  // tp_flags
+      dummy_slots,
+  };
+
+  PyObject* bases = Py_BuildValue("(O)", &PyType_Type);
+  if (!bases) return false;
+  PyObject* type = PyType_FromSpecWithBases(&dummy_spec, bases);
+  if (!type) return false;
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  Py_DECREF(bases);
+
+  bits->type_new = PyType_GetSlot((PyTypeObject*)type, Py_tp_new);
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  union {
+    // when def is msgdef, the data for this msg.
+    upb_msg* msg;
+    // when def is fielddef, owning pointer to parent
+    struct PyUpb_CMessage* parent;
+  };
+  PyObject* ext_dict;  // Weak pointer to extension dict, if any.
+  // name->obj dict for non-present msg/map/repeated, NULL if none.
+  PyUpb_WeakMap* unset_subobj_map;
+  int version;
```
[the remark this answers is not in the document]
haberman: It is: `size_t`, `void*`, so no packing opportunity here.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  return ret;
+}
+
+/*
+ * PyUpb_CMessage_LookupName()
+ *
+ * Tries to find a field or oneof named `py_name` in the message object `self`.
+ * The user can pass `f` and/or `o` to indicate whether a field or a oneof name
+ * is expected.  If the name is found and is has an expected type, the function
+ * sets `*f` or `*o` respectively and returns true.  Otherwise returns false
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  const upb_msgdef* msgdef = _PyUpb_CMessage_GetMsgdef(self);
+
+  if (!upb_msgdef_lookupname(msgdef, name, size, f, o)) {
+    if (exc_type) {
+      PyErr_Format(exc_type,
+                   "Protocol message %s has no field or oneof named %s.",
+                   upb_msgdef_fullname(msgdef), name);
+    }
+    return false;
+  } else if (!o && !*f) {
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+    if (exc_type) {
+      PyErr_Format(exc_type, "Expected a field name, but got oneof name %s.",
+                   name);
+    }
+    return false;
+  } else if (!f && !*o) {
```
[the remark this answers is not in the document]
haberman: It requires one of the two to be set. Added an assertion and updated the comment.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  }
+
+  return true;
+}
+
+static bool PyUpb_CMessage_InitMessageMapEntry(PyObject* dst, PyObject* src) {
+  if (!src || !dst) return false;
+
+  // Currently we are doing Clear()+MergeFrom().  Replace with CopyFrom() once that
+  // is implemented.
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+      if (!ok) {
+        Py_DECREF(iter);
+        return -1;
+      }
+    }
+    Py_DECREF(iter);
+  } else {
+    PyObject* tmp = PyObject_CallMethod(map, "update", "O", value);
+    if (!tmp) return -1;
+    Py_DECREF(tmp);
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+      if (!PyUpb_PyToUpb(value, f, &msgval, arena)) {
+        PyErr_Clear();
+        PyErr_Format(PyExc_ValueError, "Error initializing field %s",
+                     upb_fielddef_fullname(f));
+        return -1;
+      }
+      upb_msg_set(msg, f, msgval, arena);
+    }
+    if (PyErr_Occurred()) return -1;
+  }
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+
+  Py_DECREF(cls);
+  Py_INCREF(parent);
+  Py_INCREF(arena);
+  PyObject* ret = &msg->ob_base;
+  return ret;
+}
+
+static bool PyUpb_CMessage_IsEqual(PyUpb_CMessage* m1, PyObject* _m2) {
+  PyUpb_CMessage* m2 = (PyUpb_CMessage*)_m2;
```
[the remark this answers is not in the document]
haberman: I updated to consistently use `(void*)` for the pattern `T* self = (void*)_self`. I left the full type name for other contexts, like `foo((T*)val)`

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+  if (m1 == m2) return true;
+  if (!PyObject_TypeCheck(_m2, m1->ob_base.ob_type)) {
+    return false;
+  }
+  const upb_msgdef* m1_msgdef = _PyUpb_CMessage_GetMsgdef(m1);
+  const upb_msgdef* m2_msgdef = _PyUpb_CMessage_GetMsgdef(m2);
+  const upb_msg* m1_msg = PyUpb_CMessage_GetIfWritable((PyObject*)m1);
+  const upb_msg* m2_msg = PyUpb_CMessage_GetIfWritable(_m2);
+  (void)m2_msgdef;
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+ *   foo.bar.bar.bar.bar.bar.baz = 5
+ *
+ * This function should be called before performing any mutation of a protobuf
+ * object.
+ *
+ * Post-condition:
+ *   PyUpb_CMessage_IsUnset(self) is false
+ */
+void PyUpb_CMessage_AssureWritable(PyUpb_CMessage* self) {
+  if (!PyUpb_CMessage_IsUnset(self)) return;
```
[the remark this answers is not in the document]
haberman: I think you are correct. The only reason I favored this way was to avoid taking `O(n)` stack space, which would leave this function vulnerable to stack overflow for excessively deep messages and might therefore require introducing a recursion limit.

The code would get simpler, but the resource limit issues would get more complicated. What do you think?

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+ *   foo.MergeFrom(FooMessage(submsg={}))
+ */
+static void PyUpb_CMessage_SyncSubobjs(PyUpb_CMessage* self) {
+  PyUpb_WeakMap* subobj_map = self->unset_subobj_map;
+  upb_msg* msg = PyUpb_CMessage_GetMsg(self);
+  intptr_t iter = PYUPB_WEAKMAP_BEGIN;
+  const void* key;
+  PyObject* obj;
+
+  if (!subobj_map) return;
```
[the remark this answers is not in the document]
haberman: Done.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+
+  // The last ref to this message could disappear during iteration.
+  Py_INCREF(&self->ob_base);
```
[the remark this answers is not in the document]
haberman: Added more explanation in the comment.

python/message.c
```diff
@@ -0,0 +1,1487 @@
...
+    } else if (upb_fielddef_isseq(f)) {
+      if (!msgval.array_val) continue;
+      // TODO(haberman): re-enable when repeated fields are checked in.
+      // PyUpb_RepeatedContainer_SwitchToSet(obj, (upb_array*)msgval.array_val);
+    } else {
+      PyUpb_CMessage* sub = (PyUpb_CMessage*)obj;
+      PyUpb_ObjCache_Add(msgval.msg_val, obj);
+      assert(self == sub->parent);
+      assert(f == PyUpb_CMessage_GetFieldDef(sub));
+      Py_DECREF((PyObject*)self);
```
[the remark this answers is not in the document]
haberman: Factored into a separate function with more comments.

python/message.h
haberman: Done.

python/protobuf.c
haberman: "is" is correct. `PyObject_GetAttrString()` will yield a reference that we own. This `PyModule_AddObject()` will take ownership of that reference, keeping it alive until the module itself is destroyed.

python/protobuf.c
haberman: Done.

python/protobuf.c
haberman: Done.

haberman (commented):
I finally figured out how to batch my comments together!

PTAL.

python/message.c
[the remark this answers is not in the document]
fowles: I am playing around to find if I can see something simpler.  I have something below, but I am not 100% sure of it.  Several distinct observations:
- I think you can remove the branch on the decref by remove the final decref from after the loop
- If that is wrong you can probably add an incref above the loop and then it will work out.
- I pulled out a few helpers to make the body of the loop a bit tighter

```
static const upb_fielddef* PyUpb_CMessage_InitAsMsg(PyUpb_CMessage* m) {
  const upb_fielddef* f = PyUpb_CMessage_GetFieldDef(m);
  m->msg = upb_msg_new(upb_fielddef_msgsubdef(f), arena);
  m->def = (uintptr_t)upb_fielddef_msgsubdef(f);
  return f;
}

static void PyUpb_CMessage_SetField(PyUpb_CMessage* p, const upb_fielddef* cf, PyUpb_CMessage* c) {
    upb_msg_set(p->msg, cf, {.msg_val = c->msg}, arena);
    PyUpb_WeakMap_Delete(p->unset_subobj_map, cf);
    PyUpb_ObjCache_Add(c->msg, &c->ob_base);
}

void PyUpb_CMessage_AssureWritable(PyUpb_CMessage* child) {
  if (!PyUpb_CMessage_IsUnset(child)) return;
  // This is a non-present message. We need to create a real upb_msg for this
  // object and every parent until we reach a present message.
  upb_arena* arena = PyUpb_Arena_Get(child->arena);

  PyUpb_CMessage* next_parent = NULL;
  PyUpb_CMessage* parent = child->parent;
  const upb_fielddef* child_f = PyUpb_CMessage_InitAsMsg(child);
  do {
    const upb_fielddef* parent_f = NULL;
    if (PyUpb_CMessage_IsUnset(parent)) {
      next_parent = parent->parent;
      parent_f = PyUpb_CMessage_InitAsMsg(parent);
    }
    PyUpb_CMessage_SetField(parent, child_f, child, child_f);
    Py_DECREF(child);
    child_f = parent_f;
    child = parent;
    parent = next_parent;
  } while (PyUpb_CMessage_IsUnset(child));

  self->version++;
}
```

WDYT?

python/message.c
```diff
@@ -0,0 +1,1546 @@
...
+ * The message equivalent of PyUpb_*Container_SwitchToSet(), this transitions
+ * the wrapper from the unset state (owning a reference on self->parent) to the
+ * set state (having a non-owning pointer to self->msg).
+ */
+static void PyUpb_CMessage_SwitchToSet(PyUpb_CMessage* self,
+                                       const upb_fielddef* f, upb_msg* msg) {
+  assert(f == PyUpb_CMessage_GetFieldDef(self));
+  PyUpb_ObjCache_Add(msg, &self->ob_base);
+  Py_DECREF(&self->parent->ob_base);
+  self->msg = msg;  // Overwrites self->parent
```
fowles, line 544: it took me a really long time to understand that msg and parent were in a union (hence the comments about the overwrite).  Does C support named unions?  That might be a good way to force the code to reflect this fact explicitly.

python/message.c
```diff
@@ -0,0 +1,1546 @@
...
+                                         const upb_fielddef* field) {
+  // Non-present messages return magical "empty" messages that point to their
+  // parent, but will materialize into real messages if any fields are
+  // assigned.
+  if (!self->unset_subobj_map) {
+    self->unset_subobj_map = PyUpb_WeakMap_New();
+  }
+  PyObject* subobj = PyUpb_WeakMap_Get(self->unset_subobj_map, field);
+
+  if (!subobj) {
```
fowles, line 719: I would invert and early return here:
```
PyObject* subobj = WeakGet(...);
if (subobj) return subobj;

...
```

python/message.c
fowles: I think we have two cases
1) msg == null in which case our list should start empty
2) msg != null in which case we will have field_count items in the list

Is there some way we don't end up with one of those two?

python/message.c
fowles: can you add a comment explaining 2.  Something like

```
// Cannot fail when passed something of the correct type.
int err = PyBytes_AsStringAndSize(...);
(void)err;
assert(err >= 0);
```

python/message.c
```diff
@@ -0,0 +1,1546 @@
...
+static PyObject* PyUpb_CMessage_Clear(PyUpb_CMessage* self, PyObject* args);
+
+static PyObject* PyUpb_CMessage_ParseFromString(PyObject* self, PyObject* arg) {
+  PyObject* tmp = PyUpb_CMessage_Clear((PyUpb_CMessage*)self, NULL);
+  Py_DECREF(tmp);
+  return PyUpb_CMessage_MergeFromString(self, arg);
+}
+
+static PyObject* PyUpb_CMessage_ByteSize(PyObject* self, PyObject* args) {
+  // At the moment upb does not have a "byte size" function, so we just
```
fowles, line 996: that sounds expensive.  Can you open a bug to fix this?

python/message.c
```diff
@@ -0,0 +1,1546 @@
...
+                 upb_msgdef_fullname(msgdef), upb_fielddef_fullname(f));
+  }
+  PyUpb_CMessage_AbandonField(self, f);
+  upb_msg_clearfield(self->msg, f);
+  Py_RETURN_NONE;
+}
+
+static PyObject* PyUpb_CMessage_ClearField(PyUpb_CMessage* self,
+                                           PyObject* arg) {
+  PyUpb_CMessage_AssureWritable(self);
```
fowles, line 1041: if it is already unset, why should we ensure it is writable?  Can't we simply bail out?

python/message.c
```diff
@@ -0,0 +1,1546 @@
...
+    args = Py_BuildValue("s(OO)O", name, state->cmessage_type,
+                         state->message_class, dict);
+  } else {
+    args = Py_BuildValue("s(OOO)O", name, state->cmessage_type,
+                         state->message_class, wkt_base, dict);
+  }
+
+  PyObject* ret = cpython_bits.type_new(state->message_meta_type, args, NULL);
+  assert(Py_REFCNT(args) == 1);
+  Py_DECREF(args);
```
fowles, line 1394: random thought not even sure if this is a good idea.

I wonder if we should have a `Py_FREE` macro that expands to an assert refcount == 1 and decref.  Then we could use it consistently in places we believe we know that references will not be held and get a bit more testing on our stuff

python/protobuf.c
fowles: can you rephrase the comment to "// Reparent ownership to `m`"

python/protobuf.c
```diff
@@ -68,40 +68,121 @@ PyUpb_ModuleState *PyUpb_ModuleState_Get(void) {
...
+  }
+}
+
+bool PyUpb_WeakMap_Next(PyUpb_WeakMap *map, const void **key, PyObject **obj,
+                        intptr_t *iter) {
+  uintptr_t u_key;
+  upb_value val;
+  if (!upb_inttable_next2(&map->table, &u_key, &val, iter)) return false;
+  *key = (void *)(u_key << 3);
+  *obj = upb_value_getptr(val);
```
fowles, line 151: should this incref it or something?

python/protobuf.h
```diff
@@ -61,37 +64,78 @@ typedef struct {
...
+void PyUpb_WeakMap_Delete(PyUpb_WeakMap *map, const void *key);
+void PyUpb_WeakMap_TryDelete(PyUpb_WeakMap *map, const void *key);
 
 // Returns a new reference to an object if it exists, otherwise returns NULL.
-PyObject *PyUpb_ObjCache_Get(const void *key);
+PyObject *PyUpb_WeakMap_Get(PyUpb_WeakMap *map, const void *key);
+
+#define PYUPB_WEAKMAP_BEGIN UPB_INTTABLE_BEGIN
+
+bool PyUpb_WeakMap_Next(PyUpb_WeakMap *map, const void **key, PyObject **obj,
```
fowles, line 125: Please comment on whether this takes a new reference on the `*obj`

python/message.c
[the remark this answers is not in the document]
haberman: Your code didn't work verbatim, but I took it as inspiration. See how you like the result.

python/message.c
```diff
@@ -0,0 +1,1546 @@
...
+      Py_XDECREF(src);
+      Py_XDECREF(dst);
+      if (!ok) goto err;
+    }
+    Py_DECREF(iter);
+  } else {
+    PyObject* tmp = PyObject_CallMethod(map, "update", "O", value);
+    if (!tmp) goto err;
+  }
+  ret = 0;
```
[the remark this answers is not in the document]
fowles: Can we get asan coverage or anything?  It would be nice if we weren't relying on humans for this

python/message.c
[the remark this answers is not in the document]
fowles: That is much easier for me to follow.  Thanks for sticking with this one.

python/message.c
fowles: can you add a comment to that effect?

python/message.c
fowles: but you are already doing this with your assert?  I am fine removing the assert if that is the more correct thing...

fowles (approved):

python/message.c
haberman: Done.

PR #469, opened 2021-12-28
Run all protobuf python unit tests

haberman:
This runs the entire battery of existing unit tests for protobuf python against Python+upb. The non-passing tests have been added to an exclusion list.

The exclusion list is strict: if a test is on the exclusion list, it must fail for the test to pass overall. This will give us good ongoing visibility into which tests are still not passing.

Prior to this PR, Python+upb was only truly getting tested on my development branch. This meant we were flying blind about whether code changes made in response to PR comments were actually breaking something, since the failures would not manifest until I merged back to my development branch.

Once this PR is in, we will be in a much better position to check in the rest of the code from my Python+upb prototype, in iterative fashion, without breaking anything along the way.

This exposed a few bugs in the code. In this PR I have only fixed the bugs that were required to prevent the tests from SEGVing.


python/descriptor_containers.c
```diff
@@ -174,7 +174,6 @@ static int PyUpb_GenericSequence_IsEqual(PyUpb_GenericSequence *self,
     if (!item1 || !item2) goto error;
     int cmp = PyObject_RichCompareBool(item1, item2, Py_EQ);
     Py_DECREF(item1);
-    Py_DECREF(item2);
```
fowles, line 177: if you don't need to decref item2 here, why do you do it on line 184?

python/pb_unit_tests/BUILD
```diff
@@ -0,0 +1,61 @@
...
+      ],
+      deps = [
+          "//python:message_ext",
+          "@com_google_protobuf//:python_srcs",
+          "@com_google_protobuf//:python_common_test_protos",
+          "@com_google_protobuf//:python_specific_test_protos",
+      ],
+      legacy_create_init = False,
+  ) for test in TESTS
+]
```
fowles, line 61: I usually prefer to stamp out the parts of a build file mutliple times so that tools like buildozer can modify it more easily.  historically speaking I have always been burned by build files that contain complex logic like this.

python/descriptor_containers.c
haberman: Done.

python/pb_unit_tests/BUILD
haberman: The repetition hurts (I worry about the toil in making future manual changes), but I will try. Done.

fowles (approved):

PR #471, opened 2021-12-29
Implemented most methods on DescriptorPool

haberman:
This fixes most of the errors we were seeing in `setUp()` functions on test classes, which allows us to remove most of the test skips we were doing before.


python/descriptor.c
```diff
@@ -144,34 +145,33 @@ static PyObject* PyUpb_DescriptorBase_GetOptions(PyUpb_DescriptorBase* self,
...
-  Py_XDECREF(str);
+  PyObject* str = PyBytes_FromStringAndSize(pb, size);
   upb_arena_free(arena);
-  return ret;
+  return str;
 
 oom:
   PyErr_SetNone(PyExc_MemoryError);
-  goto done;
+  return NULL;
```
fowles, line 164: pretty sure you are missing an arena free call on the oom path

python/descriptor.c
```diff
@@ -1065,6 +1065,13 @@ static PyObject* PyUpb_FileDescriptor_GetPackage(PyObject* _self,
   return PyUnicode_FromString(upb_filedef_package(self->def));
 }
 
+static PyObject* PyUpb_FileDescriptor_GetSerializedPb(PyObject* self,
+                                                      void* closure) {
```
fowles, line 1069: do we not use the closure object?

python/descriptor_pool.c
```diff
@@ -184,40 +219,195 @@ static PyObject* PyUpb_DescriptorPool_FindExtensionByName(PyObject* _self,
...
+
+  const char* last_dot = strrchr(name, '.');
+  if (last_dot) {
+    const upb_msgdef* parent =
+        upb_symtab_lookupmsg2(self->symtab, name, last_dot - name);
+    if (parent) {
+      const upb_oneofdef* o = upb_msgdef_ntooz(parent, last_dot + 1);
+      return PyUpb_OneofDescriptor_Get(o);
+    }
+  }
```
fowles, line 317: looks like the second time you have this strchr logic.  Can you share code?

python/descriptor_pool.c
```diff
@@ -184,40 +219,195 @@ static PyObject* PyUpb_DescriptorPool_FindExtensionByName(PyObject* _self,
...
+
+static PyObject* PyUpb_DescriptorPool_FindAllExtensions(PyObject* _self,
+                                                        PyObject* msg_desc) {
+  PyUpb_DescriptorPool* self = (PyUpb_DescriptorPool*)_self;
+  const upb_msgdef* m = PyUpb_Descriptor_GetDef(msg_desc);
+  size_t n;
+  const upb_fielddef** ext = upb_symtab_getallexts(self->symtab, m, &n);
+  PyObject* ret = PyList_New(n);
+  for (size_t i = 0; i < n; i++) {
+    PyObject* field = PyUpb_FieldDescriptor_Get(ext[i]);
```
fowles, line 378: can this fail?

upb/util/def_to_proto.c
```diff
@@ -392,10 +392,14 @@ static google_protobuf_FileDescriptorProto *filedef_toproto(
   google_protobuf_FileDescriptorProto_set_name(
       proto, strviewdup(ctx, upb_filedef_name(f)));
 
-  size_t n = strlen(upb_filedef_package(f));
-  if (n) {
-    google_protobuf_FileDescriptorProto_set_package(
-        proto, strviewdup(ctx, upb_filedef_package(f)));
+  const char* package = upb_filedef_package(f);
+  size_t n;
```
fowles, line 396: the value of n from this block is not used, so I think you should embed the local into the block to keep its scope tighter

fowles (approved):

python/descriptor.c
haberman: Done.

python/descriptor.c
haberman: No we don't generally use it ever.

python/descriptor_pool.c
haberman: Done.

python/descriptor_pool.c
haberman: For OOM probably. Added a check.

upb/util/def_to_proto.c
haberman: Done. `n` is also used later in the function, but I moved the declaration down to the next usage.

PR #473, opened 2021-12-30
Added map fields for Python

haberman:


python/map.c
```diff
@@ -0,0 +1,490 @@
...
+  Py_INCREF(parent);
+  return &map->ob_base;
+}
+
+void PyUpb_MapContainer_Reify(PyObject* _self, upb_map* map) {
+  PyUpb_MapContainer* self = (PyUpb_MapContainer*)_self;
+  PyUpb_ObjCache_Add(map, &self->ob_base);
+  Py_DECREF(self->ptr.parent);
+  self->ptr.map = map;  // Overwrites self->ptr.parent.
+  self->field = self->field & ~(uintptr_t)1;
```
fowles, line 106: `self->field &= ~(uintptr_t)1`

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+  upb_arena* arena = PyUpb_Arena_Get(self->arena);
+  upb_msgval u_key, u_val;
+  if (!PyUpb_PyToUpb(key, key_f, &u_key, arena)) return -1;
+
+  if (val) {
+    if (!PyUpb_PyToUpb(val, val_f, &u_val, arena)) return -1;
+    upb_map_set(map, u_key, u_val, arena);
+  } else {
+    if (!upb_map_delete(map, u_key)) {
+      PyErr_Format(PyExc_KeyError, "Key not present in map");
```
fowles, line 149: do the other branches not need to set an error?

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+  } else {
+    Py_RETURN_FALSE;
+  }
+}
+
+PyObject* PyUpb_MapContainer_Clear(PyObject* _self, PyObject* key) {
+  PyUpb_MapContainer* self = (PyUpb_MapContainer*)_self;
+  upb_map* map = PyUpb_MapContainer_GetIfWritable(self);
+  if (map) upb_map_clear(map);
+  Py_RETURN_NONE;
```
fowles, line 176: shouldn't this do something for the non-reified case?  If feels like you should still clear something there...

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+                                        PyObject* kwargs) {
+  PyUpb_MapContainer* self = (PyUpb_MapContainer*)_self;
+  static const char* kwlist[] = {"key", "default", NULL};
+  PyObject* key;
+  PyObject* default_value = NULL;
+  upb_map* map = PyUpb_MapContainer_GetIfWritable(self);
+  if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|O", (char**)kwlist, &key,
+                                   &default_value)) {
+    return NULL;
+  }
```
fowles, line 189: after this point, don't `key` and `default_value` require decrefs?

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+  if (map && upb_map_get(map, u_key, &u_val)) {
+    return PyUpb_UpbToPy(u_val, val_f, self->arena);
+  } else {
+    if (default_value) {
+      Py_INCREF(default_value);
+      return default_value;
+    } else {
+      Py_RETURN_NONE;
+    }
+  }
```
fowles, line 207: can you unnest this control flow?
```
if (!PyUpb...) return NULL;

if (map && map_get) return ...

if (default_value) {
  inc
 retrun default_value;
}

Py_RETURN_NONE;
```

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+  PyUpb_MapContainer* self = (PyUpb_MapContainer*)_self;
+  const upb_fielddef* f = PyUpb_MapContainer_GetField(self);
+  const upb_msgdef* entry_m = upb_fielddef_msgsubdef(f);
+  return PyUpb_Descriptor_GetClass(entry_m);
+}
+
+Py_ssize_t PyUpb_MapContainer_Length(PyObject* _self) {
+  PyUpb_MapContainer* self = (PyUpb_MapContainer*)_self;
+  upb_map* map = PyUpb_MapContainer_GetIfWritable(self);
+  return map ? upb_map_size(map) : 0;
```
fowles, line 221: feels weird that the non-reified length is reported as 0

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+
+  Py_RETURN_NONE;
+}
+
+static PyObject* PyUpb_MapContainer_Repr(PyObject* _self) {
+  PyUpb_MapContainer* self = (PyUpb_MapContainer*)_self;
+  upb_map* map = PyUpb_MapContainer_GetIfWritable(self);
+  PyObject* dict = PyDict_New();
+  if (map) {
+    size_t iter = UPB_MAP_BEGIN;
```
fowles, line 257: can you move iter to be cuddled up against the loop?  I like the practice of keeping the iteration pattern very close together for easier grep and scan

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+    const upb_fielddef* f = PyUpb_MapContainer_GetField(self);
+    const upb_msgdef* entry_m = upb_fielddef_msgsubdef(f);
+    const upb_fielddef* key_f = upb_msgdef_field(entry_m, 0);
+    const upb_fielddef* val_f = upb_msgdef_field(entry_m, 1);
+    while (upb_mapiter_next(map, &iter)) {
+      PyObject* key =
+          PyUpb_UpbToPy(upb_mapiter_key(map, iter), key_f, self->arena);
+      PyObject* val =
+          PyUpb_UpbToPy(upb_mapiter_value(map, iter), val_f, self->arena);
```
fowles, line 266: can either of these fail?

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+      Py_DECREF(key);
+      Py_DECREF(val);
+    }
+  }
+  PyObject* repr = PyObject_Repr(dict);
+  Py_DECREF(dict);
+  return repr;
+}
+
+PyObject* PyUpb_MapContainer_Subscript(PyObject* _self, PyObject* key) {
```
fowles, line 277: can you move AssignSubscript and Subscript to be next to each other in the source?

python/map.c
```diff
@@ -0,0 +1,490 @@
...
+  }
+  return PyUpb_UpbToPy(u_val, val_f, self->arena);
+}
+
+PyObject* PyUpb_MapContainer_GetOrCreateWrapper(upb_map* map,
+                                                const upb_fielddef* f,
+                                                PyObject* arena) {
+  PyObject* ret = PyUpb_ObjCache_Get(map);
+
+  if (!ret) {
```
fowles, line 305: invert this conditional and early return

python/protobuf.c
```diff
@@ -254,8 +255,16 @@ static const char *PyUpb_GetClassName(PyType_Spec *spec) {
...
+  int result = PyModule_AddObject(m, name, type);
+  return type && result == 0 ? (PyTypeObject*)type : NULL;
+}
+
+PyTypeObject* PyUpb_AddClassWithBases(PyObject* m, PyType_Spec* spec,
+                                      PyObject* bases) {
+  PyObject* type = PyType_FromSpecWithBases(spec, bases);
+  const char* name = PyUpb_GetClassName(spec);
+  int result = PyModule_AddObject(m, name, type);
+  return type && result == 0 ? (PyTypeObject*)type : NULL;
```
fowles, line 267: I would rather push this cast into the variable declaration and have a simple ternary

upb/msg.c
```diff
@@ -285,6 +285,7 @@ static int _upb_mapsorter_cmpstr(const void *_a, const void *_b) {
   return a.size - b.size;
 }
 
+#include <stdio.h>
```
fowles, line 288: this feels weird

python/map.c
haberman: Done.

python/map.c
haberman: No, `PyUpb_PyToUpb()` will set the error if it fails.

python/map.c
haberman: I think you are right. Like repeated fields, I think it should `AssureWritable()` first, to reify and clear.

python/map.c
haberman: No, the local variables have "borrowed references", meaning that the refcount was not increased.

python/map.c
haberman: Done.

python/map.c
haberman: Why is that weird? A non-reified object behaves like an empty object until mutated.

python/map.c
haberman: Done.

python/map.c
haberman: Done.

python/map.c
haberman: I had been alphabetizing as the organizational principle here, but I can see the benefit of breaking that convention for this case.

python/map.c
haberman: Done.

python/protobuf.c
haberman: If the declaration is `PyObject` then other uses of `type` require a cast. I realized I needed to rework the error handling on this a bit, take a look at the new code and see what you think.

upb/msg.c
haberman: Residual printf debugging artifact, removed.

haberman:
PTAL.

python/map.c
```diff
@@ -254,16 +275,21 @@ static PyObject* PyUpb_MapContainer_Repr(PyObject* _self) {
...
+    size_t iter = UPB_MAP_BEGIN;
     while (upb_mapiter_next(map, &iter)) {
       PyObject* key =
           PyUpb_UpbToPy(upb_mapiter_key(map, iter), key_f, self->arena);
       PyObject* val =
           PyUpb_UpbToPy(upb_mapiter_value(map, iter), val_f, self->arena);
+      if (!key || !val) {
+        Py_XDECREF(key);
+        Py_XDECREF(val);
+        return NULL;
```
fowles, line 291: I think you need to `PY_DECREF(dict)` as well

python/protobuf.c
```diff
@@ -253,18 +253,24 @@ static const char *PyUpb_GetClassName(PyType_Spec *spec) {
 }
 
 PyTypeObject *PyUpb_AddClass(PyObject *m, PyType_Spec *spec) {
-  PyObject *type = PyType_FromSpec(spec);
+  PyObject *type = (void*)PyType_FromSpec(spec);
```
fowles, line 256: you added the `(void*)` here but didn't change the lhs of the object.  I am guessing you don't need this cast.

python/map.c
fowles: feels a bit like only clearing when it is not reified.  Why should "ask length, try to get the thing, ask length again" return different values for the length?

python/map.c
haberman: Done.

python/protobuf.c
haberman: Done.

haberman:
PTAL.

fowles (approved):

PR #474, opened 2021-12-31
Added ExtensionDict for Python

haberman:


python/convert.c
```diff
@@ -319,22 +319,23 @@ bool PyUpb_Array_IsEqual(const upb_array *arr1, const upb_array *arr2,
 
 bool PyUpb_Message_IsEqual(const upb_msg *msg1, const upb_msg *msg2,
                            const upb_msgdef *m) {
+  const upb_symtab* symtab = upb_filedef_symtab(upb_msgdef_file(m));
   size_t iter1 = UPB_MSG_BEGIN;
   size_t iter2 = UPB_MSG_BEGIN;
```
fowles, line 324: can you move all of this below the fast path bail outs?

python/convert.c
```diff
@@ -319,22 +319,23 @@ bool PyUpb_Array_IsEqual(const upb_array *arr1, const upb_array *arr2,
...
-    if (!ok1) break;  // Both messages are at end.
-
-    // If the two messages yielded different "next" fields, then the set of
-    // present fields is different.
-    if (f1 != f2) return false;
+  const upb_fielddef *f1, *f2;
+  upb_msgval val1, val2;
+  while (upb_msg_next(msg1, m, symtab, &f1, &val1, &iter1)) {
+    if (upb_fielddef_isextension(f1)) {
+      val2 = upb_msg_get(msg2, f1);
```
fowles, line 333: I am super confused by this case.  It feels like you fail to call `upb_msg_next` on `iter2` for this case which seems like it would lead to a bad time.

I *think* you are trying to do a parallel iteration of both of these messages and want something more like:

```
while (true) {
  bool m1_next = upb_msg_next(...1);
  bool m2_next = upb_msg_next(...2);
  if (m1_next != m2_next) return false;
  if (!m1_next) break;

  if (ismap(...) {
  } else if (isseq(....)) {
  } else {
  }
}
```

but the current behavior around `isextenstion` is confusing me greatly

python/convert.c
haberman: Done.

python/convert.c
haberman: This one definitely needed a comment, sorry about that. See if the comment helps to clarify.

This algorithm could potentially be simplified by adding a new API for iterating over extensions only. But I didn't want to slow velocity too much by going back to the drawing board for core upb APIs.

haberman:
PTAL.

python/convert.c
haberman: (btw your suggested code is almost exactly what the previous code looked like, prior to handling extensions correctly. See the red code on the left.).

fowles: so the symtab argument to `upb_msg_next` specifies whether to include extensions or not.  That is subtle and I don't love it.  I lean towards having two different names for this operation, but I have to think more on it.  How about opening a ticket to revisit this later?

fowles (approved):

python/convert.c
haberman: Yes that is fair. There are several other APIs that have similar semantics, where extensions are only recognized if a registry is provided. But I'm happy to revisit. Filed: https://github.com/protocolbuffers/upb/issues/475

PR #477, opened 2021-12-31
Fixed or accounted for nearly all errors in message_test

haberman:
Fixed/added several things:

1. We now properly raise `AttributeError` if you ask for `msg.Extensions` on a non-extendable message.
2. We now properly support the `AllowOversizeProtos` function that acts as a safeguard against messages of excessive depth.
3. We now match the text format quirk where sub-messages are printed as `msg_field { /* ... */  }`, omitting the colon that normal fields would have (`scalar_field: 123`).

Also added comments explaining why we don't pass several of the other tests in message_test.

### DIFF
--- a/python/message.c
+++ b/python/message.c
@@ -847,9 +847,12 @@ __attribute__((flatten)) static PyObject* PyUpb_CMessage_GetAttr(
   PyObject* ret = PyObject_GenericGetAttr(_self, attr);
   if (ret) return ret;
 
-  // If the attribute wasn't found, look for attributes on the class. But if a
-  // different kind of error (other than AttributeError) was found, return that.
-  if (PyErr_ExceptionMatches(PyExc_AttributeError)) {
+  // Swallow AttributeError if it occurred and try again on the metaclass
+  // to pick up class attributes.  But we have to special-case "Extensions"
+  // which affirmatively returns AttributeError when a message is not
+  // extendable.
+  if (PyErr_ExceptionMatches(PyExc_AttributeError) &&
+      strcmp(PyUpb_GetStrData(attr), "Extensions") != 0) {
     PyErr_Clear();
     return PyUpb_MessageMeta_GetAttr((PyObject*)Py_TYPE(_self), attr);
   }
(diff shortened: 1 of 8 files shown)


python/message.c
```diff
@@ -986,11 +989,14 @@ PyObject* PyUpb_CMessage_MergeFromString(PyObject* _self, PyObject* arg) {
   const upb_extreg* extreg = upb_symtab_extreg(upb_filedef_symtab(file));
   const upb_msglayout* layout = upb_msgdef_layout(msgdef);
   upb_arena* arena = PyUpb_Arena_Get(self->arena);
+  int options = 0;
+  PyUpb_ModuleState* state = PyUpb_ModuleState_Get();
+  options |=
+      UPB_DECODE_MAXDEPTH(state->allow_oversize_protos ? UINT32_MAX : 100);
```
fowles, line 995:
```
int options = UPB_DECODE...
```

seems more straightforward

python/message.c
```diff
@@ -1183,6 +1189,8 @@ PyObject* PyUpb_CMessage_SerializeInternal(PyObject* _self, PyObject* args,
   int options = 0;
   if (check_required) options |= UPB_ENCODE_CHECKREQUIRED;
   if (deterministic) options |= UPB_ENCODE_DETERMINISTIC;
+  // Python does not currently have any effective limit on serialization depth.
+  options |= UPB_ENCODE_MAXDEPTH(UINT32_MAX);
```
fowles, line 1193: move this up and do direct assignment

upb/msg.c
```diff
@@ -287,7 +287,7 @@ static int _upb_mapsorter_cmpstr(const void *_a, const void *_b) {
   _upb_mapsorter_getkeys(_a, _b, &a, &b, UPB_MAPTYPE_STRING);
   size_t common_size = UPB_MIN(a.size, b.size);
   int cmp = memcmp(a.data, b.data, common_size);
-  if (cmp) return cmp;
+  if (cmp) return -cmp;
   return a.size - b.size;
```
fowles, line 291: if `a.size` is 64-bits there is a bug on truncation to 32 bits here.

python/message.c
haberman: Done.

python/message.c
haberman: Done.

upb/msg.c
haberman: Good catch. The same applies to the other functions comparing integers. I wrote a macro that uses straightforward comparison instead of subtraction.

haberman:
PTAL.

upb/msg.c
```diff
@@ -252,45 +252,49 @@ static void _upb_mapsorter_getkeys(const void *_a, const void *_b, void *a_key,
   _upb_map_fromkey(b_tabkey, b_key, size);
 }
 
+#define COMPARE_INTEGERS(a, b) ((a) < (b) ? -1 : ((a) == (b) ? 0 : 1))
```
fowles, line 255: UPB_COMPARE_INTEGERS to avoid collisions

fowles (approved):

upb/msg.c
haberman: Done.

PR #478, opened 2022-01-02
Fixed a handful of unit tests

haberman:
Fixes:
1. Implemented "contains" for `ExtensionDict`.
2. Implemented `__deepcopy__` and `remove` for repeated fields.
3. Made all extension accessors use a common validation path for the extension descriptor.
4. Fixed some edge cases around field clearing.


python/message.c
```diff
@@ -1031,34 +1033,50 @@ static PyObject* PyUpb_CMessage_Clear(PyUpb_CMessage* self, PyObject* args) {
...
-static void PyUpb_CMessage_AbandonField(PyUpb_CMessage* self,
-                                        const upb_fielddef* f) {
-  if (self->unset_subobj_map && upb_fielddef_issubmsg(f) &&
-      !upb_fielddef_isseq(f)) {
-    PyObject* sub = PyUpb_WeakMap_Get(self->unset_subobj_map, f);
-    if (sub) {
-      PyUpb_CMessage_AssureReified((PyUpb_CMessage*)sub);
+void PyUpb_CMessage_DoClearField(PyObject* _self, const upb_fielddef* f) {
+  PyUpb_CMessage* self = (void*)_self;
+  PyUpb_CMessage_AssureReified((PyUpb_CMessage*)self);
```
fowles, line 1038: This should be `EnsureRefified` not `AssureReified` since it will actually do the work to change it if needed

python/message.c
```diff
@@ -1031,34 +1033,50 @@ static PyObject* PyUpb_CMessage_Clear(PyUpb_CMessage* self, PyObject* args) {
...
+
+  if (upb_fielddef_ismap(f)) {
+    // We have to invalidate any existing iterator over this map.
+    PyObject* obj = NULL;
+    if (self->unset_subobj_map) {
+      obj = PyUpb_WeakMap_Get(self->unset_subobj_map, f);
+    }
+    if (!obj) {
+      upb_msg* msg = PyUpb_CMessage_GetMsg(self);
```
fowles, line 1047: please add a comment here saying even if we were unable to get it ourselves, we may have a cached version that we still need to invalidate

python/repeated.c
```diff
@@ -187,9 +187,29 @@ PyObject* PyUpb_RepeatedContainer_GetOrCreateWrapper(upb_array* arr,
...
 }
 
+static PyObject* PyUpb_RepeatedContainer_MergeFrom(PyObject* _self,
+                                                   PyObject* args);
+
+PyObject* PyUpb_RepeatedContainer_DeepCopy(PyObject* _self, PyObject* value) {
+  PyUpb_RepeatedContainer* self = (PyUpb_RepeatedContainer*)_self;
+  PyUpb_RepeatedContainer* clone = (void*)PyType_GenericAlloc(Py_TYPE(_self), 0);
+  const upb_fielddef* f = PyUpb_RepeatedContainer_GetField(self);
+  if (clone == NULL) return NULL;
```
fowles, line 197: might as well move this to immediately after the alloc call

python/repeated.c
```diff
@@ -187,9 +187,29 @@ PyObject* PyUpb_RepeatedContainer_GetOrCreateWrapper(upb_array* arr,
...
+  clone->arena = PyUpb_Arena_New();
+  clone->field = (uintptr_t)PyUpb_FieldDescriptor_Get(f);
+  clone->ptr.arr =
+      upb_array_new(PyUpb_Arena_Get(clone->arena), upb_fielddef_type(f));
+  PyUpb_ObjCache_Add(clone->ptr.arr, (PyObject*)clone);
```
fowles, line 202: you cache this, but then might throw it away if the merge from fails.  That feels off to me

python/message.c
haberman: Done.

python/message.c
haberman: I realized that this code could be clarified and tested better. PTAL.

python/repeated.c
haberman: Done.

python/repeated.c
haberman: We have to do it that way, because the destructor will try to remove it from the cache, and assert-fail if it's not present there.

MergeFrom() should basically never fail, so there isn't a significant efficiency concern here.

haberman:
PTAL

fowles (approved):

PR #480, opened 2022-01-04
Fix amalgamate script to include third_party stuff

KJTsanaktsidis:
The PR https://github.com/protocolbuffers/upb/pull/429 added the utf8_range library in the third_party directory for some utf8 verification routines. However, the amalgamate script does not look there for things to squash into the amalgamation header files.

Fix this by:
    * Adding utf8_range as a lib dep for the amalgamate target in Bazel, so that build_defs.bzl will include the .c file in the amalgamation
    * Including third_party as a valid header path to squash out in the amalgamate script itself

Side note: For a bit more context, what I'm doing is embedding upb.c/h in a Ruby C extension, to generate pprof protobufs (https://github.com/google/pprof/blob/master/proto/profile.proto) as part of a performance monitoring tool I'm working on. So, after fixing this, the next thing I wanted to do was use `protoc-gen-upb` to compile the profile.proto file. 

This mostly worked, except for two things:
    * `protoc-gen-upb` puts includes in the generated files like `#include "upb/msg_internal.h"` (i.e. with `upb/` in the path), but of course after amalgamation all we have is `upb.h`.
    * `upb.h` includes `ports_def.inc` at the top, and `ports_undef.inc` at the bottom; this means that it does not actually export macros like `UPB_INLINE`, which are used by the generated `.upb.c` files.

I "fixed" this by postprocessing the generated protobufs with this piece of hackery, which seemed to work well enough:
```ruby
  port_def = File.read("#{upb_dir}/upb/port_def.inc")
  Dir["ext/ruby_memprofiler_pprof/*.upb.{c,h}"].each do |f|
    old_content = File.read(f)
    old_content.gsub!(/^\s*#include\s+["<]upb\/port_def\.inc.*$/, port_def)
    old_content.gsub!(/^\s*#include\s+["<]upb\/.*$/, '#include "upb.h"')
    File.write(f, old_content)
  end
```

If I submitted a patch to give `protoc-gen-upb` an option to generate output that plays nicely with the amalgamation, would that be of interest? How would you suggest I go about doing that?

PS: Love your comment in CONTRIBUTING.md about "if it's > 30 mins, reach out". This patch was definitely < 30 mins, but hacking on protoc-gen-upb is most certainly > 30 minutes :)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -270,6 +270,7 @@ upb_amalgamation(
         ":descriptor_upb_proto",
         ":reflection",
         ":port",
+        "//third_party/utf8_range",
     ],
 )
 
--- a/bazel/amalgamate.py
+++ b/bazel/amalgamate.py
@@ -30,6 +30,7 @@ import re
 import os
 
 INCLUDE_RE = re.compile('^#include "([^"]*)"$')
+VALID_HDR_PREFIXES = ["upb", "google", "third_party"]
 
 def parse_include(line):
   match = INCLUDE_RE.match(line)
@@ -86,7 +87,7 @@ class Amalgamator:
     include = parse_include(line)
     if not include:
       return False
-    if not (include.startswith("upb") or include.startswith("google")):
+    if not any([include.startswith(prefix) for prefix in VALID_HDR_PREFIXES]):
       return False
     if include.endswith("hpp"):
       # Skip, we don't support the amalgamation from C++.


haberman:
Thanks for the contribution, and I'm glad you liked CONTRIBUTING.md. :)

We intentionally leave third_party out of the amalgamation for policy reasons. Google OSS policy is that non-Google-authored code should always be in a `third_party` directory. This is particularly relevant for the Ruby & PHP amalgamations, which get copied into the protobuf repository (eg. https://github.com/protocolbuffers/protobuf/blob/master/ruby/ext/google/protobuf_c/ruby-upb.c).

You are using a different amalgamation, which strictly speaking wouldn't need to fulfill the same requirement. But all the amalgamations share the same `amalgamate.py`, so it appears that your change broke the Ruby/PHP amalgamations unfortunately (see the GitHub Actions errors).

With respect to `port_def.inc`/`port_undef.inc`, we really want to avoid letting user code use any of our macro definitions. To achieve that goal we need to keep `port_def.inc` and `port_undef.inc` as separate files. Also I think we want to avoid adding flags to the codegen, since that would introduce additional complexity.

Overall I would suggest, if it's not too much trouble, consider using the source files directly instead of the amalgamation. The core library doesn't require too many source files. Indeed, the amalgamation may be including more than what you need (for example, if you are dealing with a set schema, known at compile time, then you probably have no need for reflection).

If you would really prefer the amalgamation, I think the smallest we could get it down to is:

  * upb.c
  * upb.h
  * upb/port_def.inc
  * upb/port_undef.inc
  * upb/generated_deps.upb.h
  * upb/generated_deps.upbdefs.h

The last two files could be new files that we introduce. They could introduce a layer of indirection, so that the generated code only adds `#include`s for those two files. They they can `#include` either the individual `.h` files (for the regular library) or `upb.h` (for the amalgmation).

KJTsanaktsidis:
I think the github actions failures are because I didn't also include `//third_party/utf8_range` in the deps for the ruby/php amalgamate targets, so the script doesn't know to look there for includes to copy in.

I understand the policy around `third_party` & the amalgamation getting copied into the main protobuf repo, but that does leave me not quite sure how the ruby protocol buffers gem is actually going to get its hands on utf8_range. The Ruby `mkmf` module that makes the gems only builds .c files in the main extension directory; if `utf8_range.c` needs to live in a third party directory, you'll have to do some copying or something in the `extconf.rb` script I guess, or hack at the generated makefile to add a rule for `third_party/utf8_range/utf8_range.c`. At the moment it's a non-issue because it's using a verison of the amalgamation from before when utf8_range was introduced :) 

As far as "avoiding letting user code use any of our macro definitions" - I think what I would propose would be to simply embed the contents of port_def.inc/port_undef.inc into the `.upb.c`/`.upb.h` files - as long as both def and undef is coipied into the `.upb.h` file, it still own't expose the macro definitions to other includers of it.

I don't think anything is stopping me from just including a copy of the upb source tree in my gem and using its consituent parts however - in fact I was doing that before I switched to the amalgamation. That works by adding some stuff to the end of my `extconf.rb` script that hacks at the Makefile that ruby generated to build a static library target for UPB, and add it to LDFLAGS. I just thought the amalgamation seemed like a cleaner solution, but clearly it wasn't designed for this.

KJTsanaktsidis:
Well, if anybody else who wants to use upb in a ruby gem stumbles across this, I wound up just checking in a copy of upb under `ext/my_gem/vendor/upb` into my repo and slapping this at the end of my `extconf.rb`

```
File.open('Makefile', 'a') do |f|
  f.puts <<~MAKEFILE
    UPB_OBJS += $(addprefix vendor/upb/upb/,decode.o encode.o msg.o table.o upb.o)
    UPB_OBJS += vendor/upb/third_party/utf8_range/utf8_range.o
    UPB_HDRS += $(wildcard $(srcdir)/vendor/upb/upb/*.h)
    UPB_HDRS += $(wildcard $(srcdir)/vendor/upb/third_party/utf8_range/*.h)
    UPB_LIB := vendor/upb/libupb.a
    UPB_CFLAGS += -Wno-shorten-64-to-32 -Wno-sign-compare -Wno-implicit-fallthrough
    UPB_CFLAGS += -Wno-clobbered -Wno-maybe-uninitialized

    $(UPB_OBJS): CFLAGS += $(UPB_CFLAGS)
    $(UPB_OBJS): $(UPB_HDRS)
    $(UPB_OBJS): %.o : $(srcdir)/%.c
    \t$(ECHO) compiling $(<)
    \t$(Q) $(MAKEDIRS) $(@D)
    \t$(Q) $(CC) $(INCFLAGS) $(CPPFLAGS) $(CFLAGS) $(COUTFLAG)$@ -c $(CSRCFLAG)$<
    $(UPB_LIB): $(UPB_OBJS)
    \t$(ECHO) linking static library $@
    \t$(Q) $(AR) rcs $@ $^

    $(TARGET_SO): $(UPB_LIB)
    $(TARGET_SO): LOCAL_LIBS += $(UPB_LIB)
    CFLAGS += -I$(srcdir)/vendor/upb
    %.upb.o: CFLAGS += $(UPB_CFLAGS)
  MAKEFILE
```

Which is the minimum Makefile hackery I could manage to build UPB from its sources without invoking Bazel or anything like that.

haberman:
Sounds like you have a solution that doesn't require changing upb. I'll plan to close this for now.

PR #482, opened 2022-01-08
Fixed several more tests by implementing missing functionality

haberman:
This PR implements two new significant features:

* `DescriptorPool` now will try to load types lazily from a `DescriptorDatabase` if one was provided at construction time.
* `upb_symtab` will now properly unescape default values for bytes fields.

There are also a handful of other methods newly implemented:
* implemented `Descriptor.enum_values_by_name`
* implemented `EnumValueDescriptor.index`
* implemented `FieldDescriptor.extension_scope`
* implemented `OneofDescriptor.fields`


python/descriptor_pool.c
```diff
@@ -110,30 +121,63 @@ static PyObject* PyUpb_DescriptorPool_New(PyTypeObject* type, PyObject* args,
...
-    goto done;
+    return false;
   }
+  return true;
+}
+
+static PyObject* PyUpb_DescriptorPool_DoAddSerializedFile(
+    PyObject* _self, PyObject* serialized_pb) {
+  PyUpb_DescriptorPool* self = (PyUpb_DescriptorPool*)_self;
+  char* buf;
```
fowles, line 176: please move `buf` and `size` down to right before they are initialized

python/descriptor_pool.c
```diff
@@ -146,6 +190,45 @@ static PyObject* PyUpb_DescriptorPool_AddSerializedFile(
...
+    const upb_msgdef* m = PyUpb_DescriptorPool_GetFileProtoDef();
+    if (PyUpb_Message_IsEqual(proto, existing, m)) {
+      Py_INCREF(Py_None);
+      result = Py_None;
+      goto done;
+    }
+  }
+
+  if (self->db) {
+    // Load dependent files if necessary.
```
fowles, line 215: please extract this entire if into a well named helper

python/descriptor_pool.c
```diff
@@ -146,6 +190,45 @@ static PyObject* PyUpb_DescriptorPool_AddSerializedFile(
...
+    const upb_strview* deps =
+        google_protobuf_FileDescriptorProto_dependency(proto, &n);
+    for (size_t i = 0; i < n; i++) {
+      const upb_filedef* dep =
+          upb_symtab_lookupfile2(self->symtab, deps[i].data, deps[i].size);
+      if (!dep) {
+        PyObject* filename =
+            PyUnicode_FromStringAndSize(deps[i].data, deps[i].size);
+        if (!filename) goto done;
+        if (!PyUpb_DescriptorPool_TryLoadFilename(self, filename)) goto done;
```
fowles, line 226: the failure case here is missing a decref on filename

python/descriptor_pool.c
```diff
@@ -110,30 +121,63 @@ static PyObject* PyUpb_DescriptorPool_New(PyTypeObject* type, PyObject* args,
...
+                                                 PyObject* filename) {
+  if (!self->db) return false;
+  PyObject* file_proto =
+      PyObject_CallMethod(self->db, "FindFileByName", "O", filename);
+  bool ret = PyUpb_DescriptorPool_TryLoadFileProto(self, file_proto);
+  Py_XDECREF(file_proto);
+  return ret;
+}
+
+bool PyUpb_DescriptorPool_CheckNoDatabase(PyObject* _self) {
```
fowles, line 161: CheckNoDatabase has too many negatives to easily scan and "Check" is confusable with CHECK failing inside google.

Let's rename this 

```
bool PyUpb_DescriptorPool_HasDatabase(...) {
  return self->db != null;
}
```

Having now written that down, I begin to think maybe we wouldn't have a helper method here.  Maybe the 2 spots that call this should inline their test more explicitly and even provide more specialized error messages.

upb/def.c
```diff
@@ -458,6 +468,10 @@ int32_t upb_enumvaldef_number(const upb_enumvaldef *ev) {
   return ev->number;
 }
 
+uint32_t upb_enumvaldef_index(const upb_enumvaldef *ev) {
+  return ev - ev->enum_->values;
```
fowles, line 472: subtracting a pointer from something two levels deep seems suspicious...  are you sure?  If so add a comment

upb/def.c
```diff
@@ -2030,13 +2049,148 @@ static void create_oneofdef(
...
+  *ch = *ptr;
+  *src = ptr;
+  return true;
+}
+
+static char upb_symtab_TryGetHexDigit(symtab_addctx* ctx, const upb_fielddef* f,
+                                      const char** src, const char* end) {
+  char ch;
+  if (!upb_symtab_TryGetChar(src, end, &ch)) return -1;
+  if (ch >= '0' && ch <= '9') {
```
fowles, line 2071: `0 <= ch && ch <= 9`

upb/def.c
```diff
@@ -2030,13 +2049,148 @@ static void create_oneofdef(
...
+    return ch - '0';
+  }
+  ch = upb_ascii_lower(ch);
+  if (ch >= 'a' && ch <= 'f') {
```
fowles, line 2075: here too

upb/def.c
```diff
@@ -2030,13 +2049,148 @@ static void create_oneofdef(
...
+                upb_fielddef_fullname(f));
+    return 0;
+  }
+  return ret;
+}
+
+char upb_symtab_TryGetOctalDigit(const char** src, const char* end) {
+  char ch;
+  if (!upb_symtab_TryGetChar(src, end, &ch)) return -1;
+  if (ch >= '0' && ch <= '7') {
```
fowles, line 2106: `0 <= ch && ch <= 7`

upb/def.c
```diff
@@ -2030,13 +2049,148 @@ static void create_oneofdef(
...
+    return ch - '0';
+  }
+  *src -= 1;  // Char wasn't actually a octal digit.
```
fowles, line 2109: s/a octal/an octal/

upb/def.c
```diff
@@ -2030,13 +2049,148 @@ static void create_oneofdef(
...
+  return -1;
+}
+
+static char upb_symtab_ParseOctalEscape(symtab_addctx* ctx,
+                                        const upb_fielddef* f, char first,
```
fowles, line 2114: rather than pass first in, I think it is better to back up the caller by one character and do the parse of all three in here with a simple loop

upb/def.c
```diff
@@ -2030,13 +2049,148 @@ static void create_oneofdef(
...
+    case '7':
+      return upb_symtab_ParseOctalEscape(ctx, f, ch, src, end);
+  }
+  symtab_errf(ctx, "Unknown escape sequence: \\%c", ch);
+}
+
+static str_t* unescape(symtab_addctx* ctx, const upb_fielddef* f,
+                      const char* data, size_t len) {
+  // Size here is an upper bound; escape sequences could ultimately shrink it.
+  str_t* ret = symtab_alloc(ctx, sizeof(*ret) + len);
```
fowles, line 2177: this is another case where the "give back data" part of an arena could come in handy

upb/def.c
```diff
@@ -2030,13 +2049,148 @@ static void create_oneofdef(
...
+  char* dst = &ret->str[0];
+  const char* src = data;
+  const char* end = data + len;
+
+  for (; src < end; dst++, src++) {
+    if (*src == '\\') {
+      *dst = upb_symtab_ParseEscape(ctx, f, &src, end);
```
fowles, line 2184: having an helper that advances `src` and a loop that advances `src` feels error prone to me.  I would prefer to do this as:
```
while (src < end) {
  if (*src == '\\') {
    *dst++ = upb_symtab_ParseEscape(ctx, f, &src, end);
  } else {
    *dst++ = *src++;
  }
}
```

upb/def.c
```diff
@@ -2659,7 +2812,7 @@ static void create_msgdef(symtab_addctx *ctx, const char *prefix,
   m->nested_ext_count = n;
   m->nested_exts = symtab_alloc(ctx, sizeof(*m->nested_exts) * n);
   for (i = 0; i < n; i++) {
-    create_fielddef(ctx, m->full_name, NULL, fields[i], &m->nested_exts[i]);
+    create_fielddef(ctx, m->full_name, m, fields[i], &m->nested_exts[i], true);
     ((upb_fielddef*)&m->nested_exts[i])->index_ = i;
   }
```
fowles, line 2817: this method has gotten a bit too long for my taste.  Can you break it into some well named helpers?

upb/def.c
```diff
@@ -2659,7 +2812,7 @@ static void create_msgdef(symtab_addctx *ctx, const char *prefix,
   m->nested_ext_count = n;
   m->nested_exts = symtab_alloc(ctx, sizeof(*m->nested_exts) * n);
   for (i = 0; i < n; i++) {
-    create_fielddef(ctx, m->full_name, NULL, fields[i], &m->nested_exts[i]);
+    create_fielddef(ctx, m->full_name, m, fields[i], &m->nested_exts[i], true);
```
fowles, line 2815: please add `/* is_extenstion= */ true`

upb/def.c
```diff
@@ -2612,7 +2765,7 @@ static void create_msgdef(symtab_addctx *ctx, const char *prefix,
   m->field_count = n_field;
   m->fields = symtab_alloc(ctx, sizeof(*m->fields) * n_field);
   for (i = 0; i < n_field; i++) {
-    create_fielddef(ctx, m->full_name, m, fields[i], &m->fields[i]);
+    create_fielddef(ctx, m->full_name, m, fields[i], &m->fields[i], false);
```
fowles, line 2768: please add `/* is_extenstion= */ false`

upb/def.c
```diff
@@ -2972,7 +3125,8 @@ static void build_filedef(
   file->top_lvl_ext_count = n;
   file->top_lvl_exts = symtab_alloc(ctx, sizeof(*file->top_lvl_exts) * n);
   for (i = 0; i < n; i++) {
-    create_fielddef(ctx, file->package, NULL, exts[i], &file->top_lvl_exts[i]);
+    create_fielddef(ctx, file->package, NULL, exts[i], &file->top_lvl_exts[i],
+                    true);
```
fowles, line 3129: please add `/* is_extenstion= */ true`

upb/util/def_to_proto.c
```diff
@@ -92,6 +92,34 @@ static upb_strview printf_dup(upb_ToProto_Context *ctx, const char *fmt, ...) {
...
+static bool upb_isprint(char ch) {
+  return ch >= 0x20 && ch <= 0x7f;
+}
+
+static upb_strview default_bytes(upb_ToProto_Context* ctx, upb_strview val) {
+  size_t n = 0;
+  for (size_t i = 0; i < val.size; i++) {
+    n += upb_isprint(val.data[i]) ? 1 : 4;  // '\123'
+  }
+  char* p = upb_arena_malloc(ctx->arena, n);
```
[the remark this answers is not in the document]
haberman: I like this idea. There are also good use cases for *growing* the most recent allocation.  Both seem like they could potentially be accomplished within `upb_arena_realloc(arena, ptr, oldsize, newsize)` with no extra storage in the arena, if we simply compare `ptr + oldsize` with `arena->ptr`, though there could be alignment issues with this. I like the idea, but would prefer to defer.

upb/util/def_to_proto.c
```diff
@@ -92,6 +92,34 @@ static upb_strview printf_dup(upb_ToProto_Context *ctx, const char *fmt, ...) {
   return (upb_strview){.data = p, .size = n};
 }
 
+static bool upb_isprint(char ch) {
+  return ch >= 0x20 && ch <= 0x7f;
```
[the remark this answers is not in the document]
haberman: Done.

python/descriptor.c
```diff
@@ -485,6 +485,49 @@ static PyObject* PyUpb_Descriptor_GetName(PyObject* self, void* closure) {
...
+  if (!ret) return NULL;
+  int enum_count = upb_msgdef_nestedenumcount(self->def);
+  for (int i = 0; i < enum_count; i++) {
+    const upb_enumdef* e = upb_msgdef_nestedenum(self->def, i);
+    int value_count = upb_enumdef_valuecount(e);
+    for (int j = 0; j < value_count; j++) {
+      const upb_enumvaldef* ev = upb_enumdef_value(e, j);
+      const char* name = upb_enumvaldef_name(ev);
+      PyObject* val = PyLong_FromLong(upb_enumvaldef_number(ev));
+      if (!val || PyDict_SetItemString(ret, name, val) < 0) {
```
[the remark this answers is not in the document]
haberman: Done.

python/descriptor_pool.c
haberman: Done.

python/descriptor_pool.c
haberman: Done.

python/descriptor_pool.c
haberman: Done.

python/descriptor_pool.c
haberman: Done.

upb/def.c
haberman: Added comment (and added one for `upb_oneofdef_index()` too, which does the same thing). I also changed `enum_` to `parent`: hopefully this will help clarify that we are reaching one level up, not two levels deep.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

upb/def.c
haberman: Acknowledged.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

upb/def.c
haberman: Done.

haberman:
PTAL.

python/descriptor_pool.c
```diff
@@ -159,26 +159,39 @@ static bool PyUpb_DescriptorPool_TryLoadFilename(PyUpb_DescriptorPool* self,
...
-        PyExc_ValueError,
-        "Cannot call Add on a DescriptorPool that uses a DescriptorDatabase. "
-        "Add your file to the underlying database.");
-    return false;
+  return true;
+}
+
+static bool PyUpb_DescriptorPool_LoadDependentFiles(
+    PyUpb_DescriptorPool* self, google_protobuf_FileDescriptorProto* proto) {
+  // Load dependent files if necessary.
```
fowles, line 167: this comment now simply restates the method name, you can drop it

fowles (approved):

python/descriptor_pool.c
haberman: Done.

PR #484, opened 2022-01-12
Fixed a handful of reference leaks found in a debug build of Python

haberman:
Our tests contain a ref-leak detector that will activate when run under a special build of Python. This will catch any ref leaks that are triggered by our tests.

This PR fixes all ref leaks detected by our ref-leak detector. There are also a few assorted errors caught by the debug build of Python.

Ideally we can get these tests exercised in CI. But this is tricky to get hooked up in the build. On my local machine I locally built a special version of Python with debugging enabled. We would have to replicate this in GitHub actions.

Ideally this would use hermetic Python, but there is some work required to get that going.

For now I'm just fixing the leaks, and verifying with a local run of the tests.


python/message.c
```diff
@@ -100,15 +101,44 @@ static bool PyUpb_CPythonBits_Init(PyUpb_CPythonBits* bits) {
...
+  destructor subtype_dealloc = PyType_GetSlot(type, Py_tp_dealloc);
+  bits->type_dealloc = NULL;
+  for (size_t i = 0; i < 2000; i += sizeof(uintptr_t)) {
+    destructor maybe_subtype_dealloc;
+    memcpy(&maybe_subtype_dealloc, (char*)type + i, sizeof(destructor*));
+    if (maybe_subtype_dealloc == subtype_dealloc) {
+      memcpy(&bits->type_dealloc, (char*)&PyType_Type + i, sizeof(destructor*));
+      break;
+    }
+  }
```
fowles, line 127: can you extract this whole bit to a well named helper function like `upb_Pre310_PyType_GetSlot`

python/message.c
```diff
@@ -345,11 +380,19 @@ static bool PyUpb_CMessage_InitMapAttribute(PyObject* _self, PyObject* name,
...
                                                  PyObject* value) {
-  PyObject* repeated = PyUpb_CMessage_GetAttr(_self, name);
-  PyObject* tmp = PyUpb_RepeatedContainer_Extend(repeated, value);
-  if (!tmp) return false;
-  Py_DECREF(tmp);
-  return true;
+  bool ok = false;
+  PyObject* repeated = NULL;
+  PyObject* tmp = NULL;
+  repeated = PyUpb_CMessage_GetAttr(_self, name);
```
fowles, line 386:
```
PyObject* tmp = NULL;
PyObject* repeated = PyUpb_...
```

python/repeated.c
```diff
@@ -208,10 +208,12 @@ PyObject* PyUpb_RepeatedContainer_DeepCopy(PyObject* _self, PyObject* value) {
   clone->ptr.arr =
       upb_array_new(PyUpb_Arena_Get(clone->arena), upb_fielddef_type(f));
   PyUpb_ObjCache_Add(clone->ptr.arr, (PyObject*)clone);
-  if (!PyUpb_RepeatedContainer_MergeFrom((PyObject*)clone, _self)) {
+  PyObject* result = PyUpb_RepeatedContainer_MergeFrom((PyObject*)clone, _self);
+  if (!result) {
     Py_DECREF(clone);
     return NULL;
   } 
+  Py_DECREF(result);
```
fowles, line 216: do you not need to DECREF clone on this path?

fowles (approved):

python/message.c
haberman: Done.

python/message.c
haberman: Done.

python/repeated.c
haberman: We are returning it, the caller takes ownership of it.

PR #485, opened 2022-01-12
Mass API rename and clang-reformat

haberman:
Mass rename of APIs from `upb_foo_barbarbar` -> `upb_Foo_BarBarBar` style, as discussed in https://github.com/protocolbuffers/upb/issues/422

For the full list of renames so far, see `rename.sed` (which is also intended to be useful for any projects that need to migrate their code bases). Happily, `rename.sed` and `run_sed.sh` are idempotent.

Also removed a few obsolete APIs that had been previously marked for removal. There should be no functional change except for one compatibility feature I backed out (`_upb_symtab_allownameconflicts()`).

A few internal functions here and there got slightly wonky renames (eg. `_upb_array_append_accessor2` -> `_upb_Array_Append_accessor2`). We can clean these up later. The public APIs are more important, and these are looking pretty good.

There are probably some more of these renames coming, but this PR has most of the public APIs.


rename.sed
```diff
@@ -0,0 +1,494 @@
...
+#_upb_msg_map_key(const void* msg, void* key, size_t size) {
+#_upb_msg_map_value(const void* msg, void* val, size_t size) {
+#_upb_msg_map_set_value(void* msg, const void* val, size_t size) {
+#_upb_mapsorter;
+#_upb_sortedmap;
+#_upb_mapsorter_init(_upb_mapsorter *s) {
+#_upb_mapsorter_destroy(_upb_mapsorter *s) {
+#_upb_mapsorter_pushmap(_upb_mapsorter *s, upb_FieldType key_type,
+#_upb_mapsorter_popmap(_upb_mapsorter *s, _upb_sortedmap *sorted) {
+#_upb_sortedmap_next(_upb_mapsorter *s, const upb_map *map,
```
fowles, line 489: why all of this commented out stuff?

fowles (approved):

python/extension_dict.c
```diff
@@ -35,8 +35,7 @@
 // -----------------------------------------------------------------------------
 
 typedef struct {
-  PyObject_HEAD
-  PyObject* msg;  // Owning ref to our parent pessage.
+  PyObject_HEAD PyObject* msg;  // Owning ref to our parent pessage.
```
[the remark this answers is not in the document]
haberman: Added a semicolon.

rename.sed
haberman: This was a TODO list of sorts. Removed.

PR #487, opened 2022-01-17
Fixed FieldDescriptor test

haberman:
Added `TYPE_*` and `LABEL_*` constants to `FieldDescriptor. However we omit the previously-available `CPPTYPE_*` constants, as these are specific to the C++ implementation.

Also added support for finding field descriptors from a fallback database.

### DIFF
--- a/python/descriptor.c
+++ b/python/descriptor.c
@@ -1636,29 +1636,27 @@ bool PyUpb_InitDescriptor(PyObject* m) {
   }
 
   PyObject* field_desc = (PyObject*)s->descriptor_types[kPyUpb_FieldDescriptor];
-  bool ok =
-      PyUpb_SetIntAttr(field_desc, "TYPE_DOUBLE", upb_FieldType_Double) &&
-      PyUpb_SetIntAttr(field_desc, "TYPE_FLOAT", upb_FieldType_Float) &&
-      PyUpb_SetIntAttr(field_desc, "TYPE_INT64", upb_FieldType_Int64) &&
-      PyUpb_SetIntAttr(field_desc, "TYPE_UINT64", upb_FieldType_UInt64) &&
-      PyUpb_SetIntAttr(field_desc, "TYPE_INT32", upb_FieldType_Int32) &&
-      PyUpb_SetIntAttr(field_desc, "TYPE_FIXED64", upb_FieldType_Fixed64) &&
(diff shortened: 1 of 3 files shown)


python/descriptor.c
```diff
@@ -1627,5 +1635,30 @@ bool PyUpb_InitDescriptor(PyObject* m) {
...
+      PyUpb_SetIntAttr(field_desc, "TYPE_BYTES", upb_FieldType_Bytes) &&
+      PyUpb_SetIntAttr(field_desc, "TYPE_UINT32", upb_FieldType_UInt32) &&
+      PyUpb_SetIntAttr(field_desc, "TYPE_ENUM", upb_FieldType_Enum) &&
+      PyUpb_SetIntAttr(field_desc, "TYPE_SFIXED32", upb_FieldType_SFixed32) &&
+      PyUpb_SetIntAttr(field_desc, "TYPE_SFIXED64", upb_FieldType_SFixed64) &&
+      PyUpb_SetIntAttr(field_desc, "TYPE_SINT32", upb_FieldType_SInt32) &&
+      PyUpb_SetIntAttr(field_desc, "TYPE_SINT64", upb_FieldType_SInt64) &&
+      PyUpb_SetIntAttr(field_desc, "LABEL_OPTIONAL", kUpb_Label_Optional) &&
+      PyUpb_SetIntAttr(field_desc, "LABEL_REQUIRED", kUpb_Label_Required) &&
+      PyUpb_SetIntAttr(field_desc, "LABEL_REPEATED", kUpb_Label_Repeated);
```
fowles, line 1660: I can't tell if this is in a specific order or not.  In particular, uint32 is not near uint64. Can you simply sort the whole thing?

python/descriptor.c
```diff
@@ -1627,5 +1635,30 @@ bool PyUpb_InitDescriptor(PyObject* m) {
...
+  if (!ok) return false;
+
   return true;
```
fowles, line 1663: pretty sure this is equivalent to `return ok` which means you can drop the local entirely

fowles (approved):

python/descriptor.c
haberman: Done. The order is by numeric value, but it's not apparent here so I agree name is nicer.

python/descriptor.c
haberman: Done.

PR #488, opened 2022-01-17
Fixed Python test by implementing FindMethodByName

haberman:
Also fixed an improper renaming from https://github.com/protocolbuffers/upb/pull/485 by renaming `upb_DefPool_FindFileByNameforsym` -> `upb_DefPool_FindFileContainingSymbol`.


upb/def.c
```diff
@@ -1159,8 +1159,13 @@ const upb_ServiceDef* upb_DefPool_FindServiceByName(const upb_DefPool* s,
   return symtab_lookup(s, name, UPB_DEFTYPE_SERVICE);
 }
 
-const upb_FileDef* upb_DefPool_FindFileByNameforsym(const upb_DefPool* s,
-                                                    const char* name) {
+const upb_ServiceDef* upb_DefPool_FindServiceByNameWithSize(
+    const upb_DefPool* s, const char* name, size_t size) {
+  return symtab_lookup2(s, name, size, UPB_DEFTYPE_SERVICE);
```
fowles, line 1164: seems like we should have renamed symtab_lookup2 at some point...

fowles (approved):

upb/def.c
haberman: So far I've mainly focused on renames of public APIs, rather than static internal functions like `symtab_lookup2`. But I agree these are worth updating too when we get a spare moment.

PR #490, opened 2022-01-17
Fixed unit test by more rigorously testing argument type

haberman:


python/protobuf.c
```diff
@@ -302,6 +302,13 @@ const char* PyUpb_GetStrData(PyObject* obj) {
   }
 }
 
+const char* PyUpb_CheckStrData(PyObject* obj) {
```
fowles, line 305: I usually avoid "Check" as a name since it can be confused with CHECK failing in the google3 context.  What about "Verify"?

fowles (approved):

python/protobuf.c
haberman: Done.

PR #492, opened 2022-01-17
Ported protobuf's dtoa() function for text format and JSON.

haberman:
This PR changes the behavior of `str(FooMessage(abc: 2))` in Python from `abc 2.0` to `abc: 2`. The latter matches what C++ normally does.

If we go through with this, we may want to change the pure-Python implementation to drop the `.0` also, so Python+upb matches pure Python.

An alternative would be to add a flag to upb's text format encoder to always add `.0` to the end of float/double fields that would not otherwise have it.

### DIFF
--- a/upb/json_encode.c
+++ b/upb/json_encode.c
@@ -38,6 +38,7 @@
 
 #include "upb/decode.h"
 #include "upb/reflection.h"
+#include "upb/upb_internal.h"
 
 /* Must be last. */
 #include "upb/port_def.inc"
@@ -305,7 +306,7 @@ static void jsonenc_string(jsonenc* e, upb_StringView str) {
   jsonenc_putstr(e, "\"");
 }
 
-static void jsonenc_double(jsonenc* e, const char* fmt, double val) {
+static bool upb_JsonEncode_HandleSpecialDoubles(jsonenc* e, double val) {
   if (val == INFINITY) {
     jsonenc_putstr(e, "\"Infinity\"");
   } else if (val == -INFINITY) {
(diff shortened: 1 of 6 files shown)


upb/json_encode.c
```diff
@@ -305,26 +306,31 @@ static void jsonenc_string(jsonenc* e, upb_StringView str) {
...
 
-static void jsonenc_double(jsonenc* e, const char* fmt, double val) {
+static bool upb_JsonEncode_HandleSpecialDoubles(jsonenc* e, double val) {
   if (val == INFINITY) {
     jsonenc_putstr(e, "\"Infinity\"");
   } else if (val == -INFINITY) {
     jsonenc_putstr(e, "\"-Infinity\"");
   } else if (val != val) {
     jsonenc_putstr(e, "\"NaN\"");
   } else {
```
fowles, line 316: do you need to handle `-0.0` here?

upb/upb.c
```diff
@@ -326,3 +327,37 @@ bool upb_Arena_Fuse(upb_Arena* a1, upb_Arena* a2) {
...
+    assert(strtod(buf, NULL) == val);
+  }
+  upb_FixLocale(buf);
+}
+
+void _upb_EncodeRoundTripFloat(float val, char* buf, size_t size) {
+  assert(size >= kUpb_RoundTripBufferSize);
+  snprintf(buf, size, "%.*g", FLT_DIG, val);
+  if (strtof(buf, NULL) != val) {
+    snprintf(buf, size, "%.*g", FLT_DIG + 3, val);
```
fowles, line 359: why is float + 3 and dbl + 2 ?

fowles (approved):

upb/json_encode.c
haberman: It appears to be correctly handled already, added a test to verify.

upb/upb.c
haberman: Not 100% sure, this comes from the protobuf code where it has no comment. Jorg has run into this also, he may know more about the theoretical reasons for it.

PR #494, opened 2022-01-19
Fixed the build for Python 3.7

haberman:
Python 3.7 doesn't have `PyIndex_Check()` in the limited API. Calling `PyNumber_Index()` to do index coercion is compatible with older Python versions, and ends up yielding simpler code too.

CI coverage for Python 3.7 will be coming next, but first I want to submit this and unblock Ads benchmarking.


haberman:
Submitting TBR to unblock ads.

fowles (commented):
TBR Approval

PR #496, opened 2022-01-25
Fix negative hours when encode to JSON

bernardo-bruning:
This PR solved the problem around encoding data less than 1970, related in [this issue](https://github.com/protocolbuffers/protobuf/issues/9424).

The problem subdivides into two problems. The first problem is about how to use the modulus operator, the native module C is not safe for negative operations, producing the negative hours. The second problem is about precision int that introduces more 1 day when the day is less than 1970 with 3 hours.

### DIFF
--- a/tests/test_cpp.cc
+++ b/tests/test_cpp.cc
@@ -38,6 +38,8 @@
 #include "gtest/gtest.h"
 #include "tests/test_cpp.upb.h"
 #include "tests/test_cpp.upbdefs.h"
+#include "google/protobuf/timestamp.upb.h"
+#include "google/protobuf/timestamp.upbdefs.h"
 #include "upb/def.h"
 #include "upb/def.hpp"
 #include "upb/json_decode.h"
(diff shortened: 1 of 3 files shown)


upb/json_encode.c
```diff
@@ -162,20 +166,21 @@ static void jsonenc_timestamp(jsonenc* e, const upb_Message* msg,
    * Fliegel, H. F., and Van Flandern, T. C., "A Machine Algorithm for
    *   Processing Calendar Dates," Communications of the Association of
    *   Computing Machines, vol. 11 (1968), p. 657.  */
-  L = (int)(seconds / 86400) + 68569 + 2440588;
+  L = seconds / 86400.0 + 68569 + 2440588;
```
haberman, line 169: This doesn't seem correct. I think the right fix here is to make the division to always round down, instead of rounding towards zero. We can do this by ensuring that seconds is always positive.

What if we change this to something like:

```
  seconds += 62135596800;  // Ensure seconds is positive.
  L = seconds / 86400 - 719162 + 68569 + 2440588
```

If seconds is always positive, then the `safe_mod()` changes below will no longer be necessary either.

haberman (changes requested):
Could you add a test to https://github.com/protocolbuffers/upb/blob/main/tests/test_cpp.cc

We should test that a wide range of dates can round-trip.

upb/json_encode.c
bernardo-bruning: Perfect solution. I hadn't thought to keep the seconds positive.

bernardo-bruning: Perfect solution. I hadn't thought to keep the always positive.

bernardo-bruning:
HI @haberman. 

I made the tests, but I don't have practice with C and your recommendations are important for me :smiley:. I tried to introduce some fixed timestamps in tests but can create dynamically seconds to provide a wide range of dates.

bernardo-bruning:
Hi @haberman. I see that my tests are leaking memory and applied a fix to this. :sweat_smile:

tests/test_cpp.cc
```diff
@@ -148,3 +150,33 @@ TEST(Cpp, JsonNull) {
...
+    -31525200,		// 1969-01-01T03:00:00Z
+    -2208988800,	// 1900-01-01T00:00:00Z
+    -62135596800,	// 0000-01-01T00:00:00Z
+  };
+
+  for(long timestamp: timestamps) {
+    google_protobuf_Timestamp_set_seconds(timestamp_upb, timestamp);
+
+    char json[23];
+    size_t size = upb_JsonEncode(timestamp_upb, md.ptr(), NULL, 0, json, 23, NULL);
```
haberman, line 175: Please use `sizeof(json)` instead of hard-coding `23`.

Also a bigger buffer like 128 will help give some slack in case the encoding details change.

bernardo-bruning: Changed! Thanks for your suggestion.

bernardo-bruning:
@haberman, tests changed. But I forgot to mention you.

haberman (approved):

PR #499, opened 2022-02-06
Added a Bazel rule to build a binary wheel for Python

haberman:
The binary wheel contains the built `.so` files and all of the Python sources.

This is the artifact we will need to actually distribute to PyPI when Python+upb is released.

We will need to validate that this works on MacOS and Linux.

### DIFF
--- a/bazel/py_proto_library.bzl
+++ b/bazel/py_proto_library.bzl
@@ -69,9 +69,11 @@ def _generate_output_file(ctx, src, extension):
 # py_proto_library() ###########################################################
 
 def _py_proto_library_rule_impl(ctx):
-    #if len(ctx.attr.deps) != 1:
-    #    fail("only one deps dependency allowed.")
-    #dep = ctx.attr.deps[0]
+    # A real py_proto_library() should enforce this constraint.
+    # We don't bother for now, since it saves us some effort not to.
+    #
+    # if len(ctx.attr.deps) != 1:
+    #     fail("only one deps dependency allowed.")
 
     files = []
     for dep in ctx.attr.deps:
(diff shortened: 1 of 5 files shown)


bazel/py_proto_library.bzl
```diff
@@ -0,0 +1,132 @@
...
+    output_filename = paths.replace_extension(real_short_path, extension)
+    ret = ctx.actions.declare_file(output_filename)
+    return ret
+
+# py_proto_library() ###########################################################
+
+def _py_proto_library_rule_impl(ctx):
+    #if len(ctx.attr.deps) != 1:
+    #    fail("only one deps dependency allowed.")
+    #dep = ctx.attr.deps[0]
```
fowles, line 74: commented out code

python/BUILD
```diff
@@ -132,3 +145,40 @@ py_test(
...
+        "@com_google_protobuf//:wrappers_proto",
+    ],
+)
+
+py_wheel(
+    name = "binary_wheel",
+    abi = "abi3",
+    distribution = "protobuf",
+    python_tag = "cp36",
+    # TODO(haberman): we need to make this a select() that is calculated
```
fowles, line 172: can you use a tracking issue for this (unless you have a fix prepped right away already)?

python/BUILD
```diff
@@ -132,3 +145,40 @@ py_test(
...
+    # from the platform we are actually building on.
+    platform = "manylinux2014_x86_64",
+    version = "4.20.0",
+    strip_path_prefixes = ["python/"],
+    deps = [
+        ":extension_files",
+        ":well_known_proto_pb2",
+        # TODO(haberman): currently this includes the unit tests.  We should
```
fowles, line 180: here too

fowles (approved):

bazel/py_proto_library.bzl
haberman: Added a comment for why the commented-out code is there.

python/BUILD
haberman: Done.

python/BUILD
haberman: Done.

PR #501, opened 2022-02-06
Added a variant of the def-loading benchmark that builds layout dynamically

haberman:
This will help us evaluate the performance impacts (if any) that the new mini-table building code will have.

The benchmarks suggest that the mini-table building in upb is a very small portion of the overall cost of building defs. We see the same result in proto2, where building DynamicMessage layouts appears to be a small cost compared with loading a descriptor into a descriptor pool.

### DIFF
--- a/benchmarks/benchmark.cc
+++ b/benchmarks/benchmark.cc
@@ -33,6 +33,7 @@
 #include "benchmarks/descriptor_sv.pb.h"
 #include "google/ads/googleads/v7/services/google_ads_service.upbdefs.h"
 #include "google/protobuf/descriptor.pb.h"
+#include "google/protobuf/dynamic_message.h"
 #include "upb/def.hpp"
 
 upb_StringView descriptor = benchmarks_descriptor_proto_upbdefinit.descriptor;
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark.cc
```diff
@@ -70,47 +71,90 @@ static void BM_ArenaInitialBlockOneAlloc(benchmark::State& state) {
...
-    protobuf::StringPiece input(descriptor.data, descriptor.size);
-    auto proto =
-        protobuf::Arena::CreateMessage<protobuf::FileDescriptorProto>(&arena);
-    protobuf::DescriptorPool pool;
-    bool ok = proto->ParseFrom<protobuf::MessageLite::kMergePartial>(input) &&
-              pool.BuildFile(*proto) != nullptr;
-    if (!ok) {
-      printf("Failed to add file.\n");
-      exit(1);
+    if (Mode == NoLayout) {
```
fowles, line 140: is this boolean backwards?  It feels like you should call NoLayout when NoLayout is true

fowles (approved):

benchmarks/benchmark.cc
haberman: I can see what that is confusing. I've renamed the function to `LoadDefInit_BuildLayout()` to make this more clear. The function forces a layout to be computed by not passing in the compiled-in layout. It was called `NoLayout` because it does not pass a layout in (forcing one to be computed), whereas the `NoLayout` benchmarks are benchmarking the case that a layout is not computed.

PR #515, opened 2022-02-20
force explicit conversions to desired types in generated code

fowles:


upb/upb.h
```diff
@@ -90,6 +90,18 @@ UPB_INLINE bool upb_StringView_IsEqual(upb_StringView a, upb_StringView b) {
   return a.size == b.size && memcmp(a.data, b.data, a.size) == 0;
 }
 
+/** upb_*Int* conversion routines ********************************************/
+
+UPB_INLINE int32_t upb_Int32_FromI(int v) { return (int32_t)v; }
```
haberman, line 95: As these are not intended for users, can we put them in `upb/msg_internal.h`?

That file is growing a bit too large, but it's the place for stuff used by generated code and not intended for users.

fowles: Done

haberman (approved):

PR #516, opened 2022-02-20
explicitly include deps for internal headers

fowles:


haberman:
I think we might want to make a separate rule for the internal library, but this is fine for now.

haberman (approved):

PR #525, opened 2022-02-23
Fixed two bugs with proto2 enums

haberman:
1. The mask was getting improperly truncated for values 32-63.
2. We were not handling duplicated enum values.

In both cases asserts would catch the error. However in no-asserts builds, this bug could cause values parsed from the wire to be inappropriate placed into the unknown field set.

### DIFF
--- a/upb/def.c
+++ b/upb/def.c
@@ -2397,6 +2397,12 @@ static int count_bits_debug(uint64_t x) {
   return n;
 }
 
+static int compare_int32(const void* a_ptr, const void* b_ptr) {
+  int32_t a = *(int32_t*)a_ptr;
+  int32_t b = *(int32_t*)b_ptr;
+  return ((a) < (b) ? -1 : ((a) == (b) ? 0 : 1));
+}
+
 upb_MiniTable_Enum* create_enumlayout(symtab_addctx* ctx,
                                       const upb_EnumDef* e) {
   int n = 0;
@@ -2426,6 +2432,17 @@ upb_MiniTable_Enum* create_enumlayout(symtab_addctx* ctx,
     }
     UPB_ASSERT(p == values + n);
   }
+
+  // Enums can have duplicate values; we must sort+uniq them.
+  qsort(values, n, sizeof(*values), &compare_int32);
+
+  int dst = 0;
+  for (int i = 0; i < n; dst++) {
+    int32_t val = values[i];
+    while (i < n && values[i] == val) i++;  // Skip duplicates.
+    values[dst] = val;
+  }
+  n = dst;
 
   UPB_ASSERT(upb_inttable_count(&e->iton) == n + count_bits_debug(mask));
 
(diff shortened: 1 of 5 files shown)


upb/def.c
```diff
@@ -2397,6 +2397,12 @@ static int count_bits_debug(uint64_t x) {
   return n;
 }
 
+static int compare_int32(const void* a_ptr, const void* b_ptr) {
+  int32_t a = *(int32_t*)a_ptr;
+  int32_t b = *(int32_t*)b_ptr;
+  return ((a) < (b) ? -1 : ((a) == (b) ? 0 : 1));
```
fowles, line 2403: you don't need all the extra () now

upb/def.h
```diff
@@ -389,7 +389,15 @@ typedef struct _upb_DefPool_Init {
   upb_StringView descriptor; /* Serialized descriptor. */
 } _upb_DefPool_Init;
 
-bool _upb_DefPool_LoadDefInit(upb_DefPool* s, const _upb_DefPool_Init* init);
+// Should only be directly called by tests.  This varant lets us suppress
```
fowles, line 392: s/varant/variant/

upbc/protoc-gen-upb.cc
```diff
@@ -95,6 +95,17 @@ std::vector<const protobuf::EnumDescriptor*> SortedEnums(
...
 }
 
+std::vector<int32_t> SortedUniqueEnumNumbers(
+    const protobuf::EnumDescriptor* e) {
+  std::vector<int32_t> values;
+  for (int i = 0; i < e->value_count(); i++) {
+    values.push_back(e->value(i)->number());
+  }
+  std::sort(values.begin(), values.end());
+  std::unique(values.begin(), values.end());
```
fowles, line 105: don't you need to truncate after the unique call?

haberman: Good catch. I was redundantly achieving the same by using a `flat_hash_set` in the surrounding function. I've added an `erase()` here and removed the redundant code in the caller.

haberman:
PTAL.

upb/def.c
```diff
@@ -2402,7 +2402,7 @@ static int count_bits_debug(uint64_t x) {
 static int compare_int32(const void* a_ptr, const void* b_ptr) {
   int32_t a = *(int32_t*)a_ptr;
   int32_t b = *(int32_t*)b_ptr;
-  return ((a) < (b) ? -1 : ((a) == (b) ? 0 : 1));
+  return (a) < (b) ? -1 : ((a) == (b) ? 0 : 1);
```
fowles, line 2406: still have parens you don't need around `(a)` and `(b)`

upbc/protoc-gen-upb.cc
```diff
@@ -1311,22 +1312,20 @@ int WriteEnums(const protobuf::FileDescriptor* file, Output& output) {
...
-    std::sort(values_vec.begin(), values_vec.end());
 
-    if (!values_vec.empty()) {
+    if (!values.empty()) {
       values_init = EnumInit(e) + "_values";
       output("static const int32_t $0[$1] = {\n", values_init,
-             values_vec.size());
-      for (auto value : values_vec) {
+             values.size());
+      for (auto value : values) {
```
fowles, line 1327: bear auto always makes me worry about expensive copies.  can you name the type

fowles (approved):

PR #526, opened 2022-02-24
Added fuzzing of symtab build, and fixed a handful of minor bugs.

haberman:
This extends our existing fuzz test to cover not only the parser, but the code that loads a descriptor into a `upb_DefPool`.

I expect this would have caught https://github.com/protocolbuffers/upb/pull/525

### DIFF
--- a/upb/def.c
+++ b/upb/def.c
@@ -1601,6 +1601,10 @@ static void make_layout(symtab_addctx* ctx, const upb_MessageDef* m) {
 
     if (upb_OneofDef_IsSynthetic(o)) continue;
 
+    if (o->field_count == 0) {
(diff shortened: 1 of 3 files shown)


upb/def.c
```diff
@@ -2949,7 +2958,7 @@ static void build_filedef(
   int32_t* mutable_weak_deps = (int32_t*)file->weak_deps;
   for (i = 0; i < n; i++) {
     if (weak_deps[i] >= file->dep_count) {
-      symtab_errf(ctx, "public_dep %d is out of range", (int)public_deps[i]);
+      symtab_errf(ctx, "public_dep %d is out of range", (int)weak_deps[i]);
```
fowles, line 2961: the error says "public_dep" but the print is a weak_dep

fowles (approved):

PR #551, opened 2022-03-14
Ruby needs the "seed" parameter in our hash function.

haberman:
This fixes a mistake in https://github.com/protocolbuffers/upb/pull/545. When exposing the hash function for Ruby to use, I forgot that Ruby needs to be able to pass the `seed` parameter.

### DIFF
--- a/upb/table.c
+++ b/upb/table.c
@@ -433,7 +433,7 @@ const uint64_t kWyhashSalt[5] = {
     0x082EFA98EC4E6C89ULL, 0x452821E638D01377ULL,
 };
 
-uint32_t _upb_Hash(const char* p, size_t n, uint64_t seed) {
+uint32_t _upb_Hash(const void* p, size_t n, uint64_t seed) {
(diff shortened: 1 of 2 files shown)


upb/table.c
```diff
@@ -496,20 +500,20 @@ bool upb_strtable_insert(upb_strtable* t, const char* k, size_t len,
   tabkey = strcopy(key, a);
   if (tabkey == 0) return false;
 
-  hash = _upb_Hash(key.str.str, key.str.len);
+  hash = _upb_Hash_NoSeed(key.str.str, key.str.len);
```
fowles, line 503: why did this go to NoSeed and not use wyhashseed?

haberman: Wyhash has both "seed" and "salt". We still use Wyhash's salt, but the seed is generally only useful when hashing multiple things and wanting to propagate state from one call to the next.

haberman:
PTAL.

fowles (approved):

PR #567, opened 2022-04-02
Release Builds

deannagarcia:
Adding bazel targets for release

Required flags:
* `--action_env MACOSX_DEPLOYMENT_TARGET=10.9` just for the macos target.


python/dist/BUILD
```diff
@@ -0,0 +1,155 @@
...
+)
+
+config_setting(
+    name = "win64_cpu",
+    values = {"cpu": "win64"},
+)
+
+py_wheel(
+    name = "binary_wheel",
+    abi = "abi3",
```
haberman, line 92: abi3 is only appropriate for the limited API. For the full API we need to use something like "cp39", "cp38", etc.

python/dist/BUILD
```diff
@@ -0,0 +1,155 @@
...
+    distribution = "protobuf",
+    # TODO(https://github.com/protocolbuffers/upb/issues/502): we need to make
```
haberman, line 94: You can remove this as you are fixing the bug. :)

python/dist/BUILD
```diff
@@ -0,0 +1,155 @@
...
+    # this a select() that is calculated from the platform we are actually
+    # building on.
+    platform = select({
+        ":x86_64_cpu": "manylinux2014_x86_64",
+        ":aarch64_cpu": "manylinux2014_aarch64",
+        ":osx-x86_64_cpu": "macosx_12_1_universal",
```
haberman, line 100: I think we'll want to go for 10_9 here to match our existing release: https://pypi.org/project/protobuf/#files

I think this may involve setting MACOSX_DEPLOYMENT_TARGET=10.9. I only know a little about this flag so you'll want to double-check me on that.

haberman (commented):
This is looking good! To answer your questions:

> How do I test to ensure that the wheels are correct?

`pip install` the wheel and see if you can run the tests. Ideally the test could create a virtualenv so that `pip install` doesn't affect the environment.

> Do I have the correct dependencies in the py_wheel rule?

Yes, it's still including the tests unnecessarily, but honestly it might actually be nice to ship the tests so it's easier to run the tests after installing the package.

> String escaping in the workspace file was hard for me, I ended up needing the flag --incompatible_restrict_string_escapes=false, but hoping to find a better way to do this in the future.

Where did this show up? It's not immediately obvious to me.

It might make sense to split some of the WORKSPACE logic into repository rules, instead of doing eg. interpolation right in WORKSPACE.

bazel/protobuf.patch
```diff
@@ -87,3 +87,80 @@
...
++)
++
+ cc_toolchain_config(
+     name = "linux-aarch_64-config",
+     sysroot = "/opt/manylinux/2014/aarch64",
+@@ -85,6 +85,7 @@ cc_toolchain_config(
+         "-I/usr/tools/apple_sdks/xcode_13_0/macosx/usr/include/c++/v1",
+         "-I/usr/tools/apple_sdks/xcode_13_0/macosx/usr/include"
+     ],
++    extra_linker_flags = ["-undefined", "dynamic_lookup"],
```
haberman, line 120: `-undefined dynamic_lookup` should go directly onto the Python `cc_binary(linkshared=1)` rule, not on the toolchain.

We don't want `-undefined dynamic_lookup` be used on regular binaries. Regular binaries should throw link errors when a symbol is not available, otherwise the binary will fail at runtime:

Good behavior, this is what we want:
```
$ cat foo.c
int f();

int main() {
  return f();
}
$ gcc -o foo foo.c
Undefined symbols for architecture x86_64:
  "_f", referenced from:
      _main in foo-db1b0c.o
ld: symbol(s) not found for architecture x86_64
```

Bad behavior, the link succeeds but the binary fails at runtime.
```
$ gcc -o foo foo.c -undefined dynamic_lookup
(base) √ Joshs-MacBook-Pro ~ $ ./foo 
dyld[46235]: missing symbol called
zsh: abort      ./foo
```

bazel/py_extension.bzl
```diff
@@ -12,27 +12,34 @@ def py_extension(name, srcs, deps = []):
...
+
     native.cc_binary(
         name = name,
         srcs = srcs,
-        copts = UPB_DEFAULT_COPTS + [
+        copts = UPB_DEFAULT_COPTS + select(LIMITED_API_FLAG_SELECT) + [
             # The Python API requires patterns that are ISO C incompatible, like
             # casts between function pointers and object pointers.
             "-Wno-pedantic",
+            # "--compat-implib",
```
haberman, line 28: We can remove this, it was a troubleshooting thing that didn't work out.

python/dist/BUILD
deannagarcia: Added a select to fix.

python/dist/BUILD
deannagarcia: Removed.

python/dist/BUILD
deannagarcia: Edited the name, but leaving this open to talk more about the MACOSX_DEPLOYMENT_TARGET. 

I found a place in the APK files where this variable is set, should I change it there or is there a compiler flag I can change?

bazel/protobuf.patch
deannagarcia: Moved!

bazel/python_downloads.bzl
```diff
@@ -0,0 +1,63 @@
...
+def limited_api_download(version, sha256):
+    http_archive(
+        name = "python-{}".format(version),
+        urls = [
+            "https://www.python.org/ftp/python/{0}/Python-{0}.tgz"
+            .format(version)],
+        sha256 = sha256,
+        build_file_content = limited_api_build_file.format(version),
+        patch_cmds = [
+            "echo \#define SIZEOF_WCHAR_T 4 > Python-{}/Include/pyconfig.h"
```
deannagarcia, line 21: This is where I'm using string escaping to get `#define`. Without escaping this produces an error.

deannagarcia:
I can't figure out how to use transitions here. The problem is that `py_wheel` can't take a transition. Since it's a rule we can't make a wrapper rule that could accept the transition. Any ideas on how to do this?

Otherwise I think the only option is to make a `.bazelrc` file and make an individual config for each build and do it similarly to how protoc is built now.

python/dist/BUILD
```diff
@@ -0,0 +1,154 @@
...
+    extension = "//python:_message",
+)
+
+py_dist_module(
+    name = "api_implementation_mod",
+    module_name = "google.protobuf.internal.api_implementation",
+    extension = "//python:_api_implementation",
+)
+
+# py_proto_library(
```
deannagarcia, line 48: Do we need this `py_proto_library` anywhere?

python/dist/BUILD
haberman: We should set `MACOSX_DEPLOYMENT_TARGET=10.9` as an environment variable.

Here is some info I found: https://stackoverflow.com/a/33308902

Looks like the easiest way to set it is via `--action_env`: https://stackoverflow.com/a/57902656.

haberman:
> I can't figure out how to use transitions here. The problem is that py_wheel can't take a transition. Since it's a rule we can't make a wrapper rule that could accept the transition. Any ideas on how to do this?

Can't we make it an outgoing transition on `py_dist()` (ie as it is currently written?)

Then `py_wheel()` isn't aware that there is any transition. It is just instantiated several times, with different config combinations that came from `py_dist()`.

If `py_wheel()` is working now, we should commit what we have and worry about the transitions in a second PR.

python/dist/BUILD
deannagarcia: Ok that should work! I think that will fit into the transitions part of the work, so I will leave this as is and put a note in the description.

python/dist/BUILD
deannagarcia: Added.

deannagarcia:
I think this is ready for a final review! Transitions still aren't in, but that can come in a follow-up PR.

bazel/python_downloads.bzl
haberman: I was able to get this to work with:

```
        patch_cmds = [
            "echo '#define SIZEOF_WCHAR_T 4' > Python-{}/Include/pyconfig.h"
            .format(version)]
```

haberman:
I tried a few examples to see if I got outputs matching the scheme in the doc.

```
$ bazel build python/dist:api_implementation_mod_file --//python:limited_api=true --//python:python_version=37
Target //python/dist:api_implementation_mod_file up-to-date:
  bazel-bin/python/dist/google/protobuf/internal/api_implementation.cpython-37-x86_64-linux-gnu.so
```

I think that should be `api_implementation.abi3.so`

```
$ bazel build python/dist:api_implementation_mod_file --//python:limited_api=false --//python:python_version=37Target //python/dist:api_implementation_mod_file up-to-date:
  bazel-bin/python/dist/google/protobuf/internal/api_implementation.so
```

I think that should be `api_implementation.cpython-37m-x86_64-linux-gnu.so`.

bazel/python_downloads.bzl
deannagarcia: Changed to that!

deannagarcia:
Fixed the tests that were failing and merged with main. 

The bazel tests are failing because they don't have the proper configs. How should we fix that?

haberman:
> The bazel tests are failing because they don't have the proper configs. How should we fix that?

`//python:python_version` should default to `system`, which should depend on `@system_python//:python_headers` like we do now.

haberman:
To make the tests pass, does it work to add `tags = ["manual"]` to the `py_dist()` rule?

haberman:
Merged in https://github.com/protocolbuffers/upb/commit/ee6b1abb35ed852201937402a65be4b866849fec

PR #637, opened 2022-06-01
Update BUILD

aranguyen:
Update BUILD to use constraints values from @platforms instead of @bazel_tools//platforms https://github.com/bazelbuild/bazel/issues/8622


haberman:
Obsolete PR, this was submitted in https://github.com/protocolbuffers/upb/commit/e5f26018368b11aab672e8e8bb76513f3620c579

PR #646, opened 2022-06-08
Add numpy install to python_tests.yaml

zhangskz:


zhangskz:
Install seems to have worked: https://github.com/protocolbuffers/upb/runs/6797956278?check_suite_focus=true

PR #677, opened 2022-06-28
Java wrapper for upb

cgrushko:
This shows how to use upb from Java.
I don't expect this to be merged.


acozzette:
It sounds like this PR is just a proof-of-concept demonstration, so let me go ahead and close it. We recently moved upb to the main protobuf [repo](https://github.com/protocolbuffers/protobuf).

PR #726, opened 2022-08-05
Fixed comparison of maps in Python.

haberman:
The `__eq__` method from our base class was not getting called, because we defined `tp_hash` as `NotImplemented`.  Per [the docs on `tp_hash`](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_hash), `tp_hash` and `tp_richcompare` are inherited together, so defining `tp_hash` was inhibiting the inheritance of `tp_richcompare`.

PiperOrigin-RevId: 465184018


haberman:
This is a cherry-pick of https://github.com/protocolbuffers/upb/pull/725

fowles (approved):

PR #729, opened 2022-08-10
Add upb_Arena_SpaceAllocated function

casperisfine:
Ref: https://github.com/protocolbuffers/protobuf/pull/10291

Ruby types defined though native extensions should register
a function that report their memory footprint in bytes.

This feature is used by various memory profiling tools.

As suggested by @haberman.

NB: the implementation is a bit simpler than what we had in https://github.com/protocolbuffers/protobuf/pull/10291, here we return the arena allocated size, but we don't care about fused arenas. I think dividing by the number of fused arena is an implementation detail of the Ruby extension.

### DIFF
--- a/upb/arena.c
+++ b/upb/arena.c
@@ -69,6 +69,24 @@ static upb_Arena* arena_findroot(upb_Arena* a) {
   return a;
 }
 
+size_t upb_Arena_SpaceAllocated(upb_Arena* arena) {
+  arena = arena_findroot(arena);
+  size_t memsize = 0;
+
+  mem_block* block = arena->freelist;
+
+  while (block) {
+    memsize += sizeof(mem_block) + block->size;
+    block = block->next;
+  }
+
+  return memsize;
+}
+
+uint32_t upb_Arena_DebugRefCount(upb_Arena* arena) {
+  return arena_findroot(arena)->refcount;
+}
+
 static void upb_Arena_addblock(upb_Arena* a, upb_Arena* root, void* ptr,
                                size_t size) {
   mem_block* block = ptr;
--- a/upb/arena.h
+++ b/upb/arena.h
@@ -72,6 +72,8 @@ void upb_Arena_Free(upb_Arena* a);
 bool upb_Arena_AddCleanup(upb_Arena* a, void* ud, upb_CleanupFunc* func);
 bool upb_Arena_Fuse(upb_Arena* a, upb_Arena* b);
 void* _upb_Arena_SlowMalloc(upb_Arena* a, size_t size);
+size_t upb_Arena_SpaceAllocated(upb_Arena* arena);
+uint32_t upb_Arena_DebugRefCount(upb_Arena* arena);
 
 UPB_INLINE upb_alloc* upb_Arena_Alloc(upb_Arena* a) { return (upb_alloc*)a; }
 


haberman:
This looks good. We should also expose a `upb_Arena_DebugRefCount()` that just returns the ref count. The members of `upb_Arena` should be private from the perspective of the Ruby extension. Putting `Debug` in the name will make clear that the return is for informational purposes only, and shouldn't be used to control any logic.

casperisfine:
> We should also expose a `upb_Arena_DebugRefCount()` that just returns the ref count.

Done.

upb/arena.h
```diff
@@ -80,6 +81,10 @@ UPB_INLINE size_t _upb_ArenaHas(upb_Arena* a) {
   return (size_t)(h->end - h->ptr);
 }
 
+UPB_INLINE uint_32_t upb_Arena_DebugRefCount(upb_Arena *arena) {
+  return arena->refcount;
```
haberman, line 85: We need to find the group leader first. The correct algorithm here would be:

uint_32_t upb_Arena_DebugRefCount(upb_Arena *arena) {
  return arena_findroot(arena)->refcount;
}

casperisfine: Indeed. I also forgot it in `upb_Arena_SpaceAllocated`. I added both.

upb/arena.c
```diff
@@ -69,6 +69,24 @@ static upb_Arena* arena_findroot(upb_Arena* a) {
...
+
+  while (block) {
+    memsize += sizeof(mem_block) + block->size;
+    block = block->next;
+  }
+
+  return memsize;
+}
+
+uint_32_t upb_Arena_DebugRefCount(upb_Arena *arena) {
```
haberman, line 86: Sorry, I mistyped before, this should be just `uint32_t` (not `uint_32_t`).

casperisfine: No worries, I should have caught it.

upb/arena.h
```diff
@@ -72,6 +72,8 @@ void upb_Arena_Free(upb_Arena* a);
 bool upb_Arena_AddCleanup(upb_Arena* a, void* ud, upb_CleanupFunc* func);
 bool upb_Arena_Fuse(upb_Arena* a, upb_Arena* b);
 void* _upb_Arena_SlowMalloc(upb_Arena* a, size_t size);
+size_t upb_Arena_SpaceAllocated(upb_Arena *arena);
+uint_32_t upb_Arena_DebugRefCount(upb_Arena *arena);
```
haberman, line 76: Same issue here.

Are you able to run the tests? If you install Bazel you should be able to run:

```
$ bazel test :all
```

casperisfine:
@haberman apologies for not double checking, should be good to go now.

haberman:
Would you be able to pull the latest changes from `main`? I've fixed the GitHub Actions to work properly for external contributors. Sorry for the hassle on this CL, we were still in the process of getting our CI system healthy.

upb/arena.c
```diff
@@ -69,6 +69,24 @@ static upb_Arena* arena_findroot(upb_Arena* a) {
...
+  while (block) {
+    memsize += sizeof(mem_block) + block->size;
+    block = block->next;
+  }
+
+  return memsize;
+}
+
+uint32_t upb_Arena_DebugRefCount(upb_Arena *arena) {
+    return arena_findroot(arena)->refcount;
```
haberman, line 87: This should be indented 2, not 4. If you have clang-format, running it from the base directory should fix this.

casperisfine:
@haberman rebased

casperisfine:
Thanks!

PR #745, opened 2022-08-24
Remove patter that does not glob anything

limdor:
This pattern does not glob anything because inside the upbc folder there is a BUILD file.
Removing this allows to build upb with the flag incompatible_disallow_empty_glob

Note: This is needed in order to build Bazel with the flag flipped.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -963,7 +963,6 @@ exports_files(
 filegroup(
     name = "cmake_files",
     srcs = glob([
-        "upbc/**/*",
         "upb/**/*",
         "third_party/**/*",
     ]),


ericsalo (commented):
This seems like an innocent enough change but many of the tests are now breaking.

limdor:
> This seems like an innocent enough change but many of the tests are now breaking.

Can you point me to the log of a failing test? The ones here in github they seem to fail but to me looks like something unrelated. Or do you think that is related to the change?

> Error: google-github-actions/auth failed with: retry function failed with 0 attempts: the GitHub Action workflow must specify exactly one of "workload_identity_provider" or "credentials_json"! If you are specifying input values via GitHub secrets, ensure the secret is being injected into the environment. By default, secrets are not passed to workflows triggered from forks, including Dependabot.

Another option is to add the allow_empty=True, but it would be good to investigate why the tests are failing, I cannot see how this could change behaviour.

ericsalo:
The cause of the failures has been identified. Once we have fixed them on our end we should be able to proceed with this PR. Unfortunately most of the team will be away on an offsite this coming week so it may be a little while before we can get to it.

limdor:
Ok, thanks for the update. The goal would be before the cut for Bazel 6 on September 12th to have this merged and Bazel updated with a version that contains this change.

limdor:
@ericsalo any updates on this?

limdor:
@ericsalo would it be possible to bring this change to the 21.x branch? What would be the process for that?

fowles:
a PR with a cherrypick would suffice

limdor:
Thanks @fowles, I created the PR in https://github.com/protocolbuffers/upb/pull/780

PR #764, opened 2022-09-03
Updated staleness test and amalgamator to work cross-repo.

haberman:
This is to support https://github.com/protocolbuffers/protobuf/pull/10496


bazel/amalgamate.py
```diff
@@ -35,95 +35,97 @@ def parse_include(line):
   match = INCLUDE_RE.match(line)
   return match.groups()[0] if match else None
 
+def is_core_upb_header(fname):
```
mkruskal-google, line 38: This doesn't look like it's used anywhere?

mkruskal-google (approved):

bazel/amalgamate.py
haberman: Done.

PR #771, opened 2022-09-08
Refactor generic comparison functions into upb/util/compare

cmburn:
Allows other bindings to use them without writing everything twice


haberman:
For what it's worth, I think this is a good direction and something we will want to do. Though it may require a bit of iteration to nail down the semantics.

Any particular reason you closed this?

cmburn:
I... have no idea why I closed this, I thought it was still open to be honest. Should close this again I make a new one that's up-to-date with upstream instead?

PR #781, opened 2022-09-12
use constraints values in @platforms instead of @bazel_tools//platfor…

limdor:
…ms https://github.com/bazelbuild/bazel/issues/8622

PiperOrigin-RevId: 452387904
(cherry picked from commit e5f26018368b11aab672e8e8bb76513f3620c579)

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -58,7 +58,7 @@ exports_files(
 
 config_setting(
     name = "windows",
-    constraint_values = ["@bazel_tools//platforms:windows"],
+    constraint_values = ["@platforms//os:windows"],
     visibility = ["//visibility:public"],
 )
 


limdor:
I guess that the CLA failure is because the initial author is the Prorobuf Team.

limdor:
@haberman @ericsalo could any of you support with this cherry pick? Would it be fine if I take the changes manually without a cherry pick?

haberman (approved):

limdor:
Thanks for approving, will you take care also of merging. It would be cool if could be done soon with this way we can bump the version from Bazel before Bazel 6.

limdor:
Thanks!!

Issue #784, opened 2022-09-14
upb is broken with Bazel@HEAD


meteorcloudy:
First reported at https://github.com/bazelbuild/continuous-integration/issues/1439

https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2631#01833787-115d-48d9-938c-084b2f36562c

```
5:46:27) ERROR: /var/lib/buildkite-agent/builds/bk-docker-xlpj/bazel-downstream-projects/upb/upb/bindings/lua/BUILD.bazel:96:14: in //upb/bindings/lua:lua_proto_library.bzl%_lua_proto_library_aspect aspect on proto_library rule //upb/bindings/lua:test_proto:
Traceback (most recent call last):
	File "/var/lib/buildkite-agent/builds/bk-docker-xlpj/bazel-downstream-projects/upb/upb/bindings/lua/lua_proto_library.bzl", line 118, column 32, in _lua_proto_library_aspect_impl
		files = _compile_upb_protos(ctx, proto_info, proto_info.direct_sources)
	File "/var/lib/buildkite-agent/builds/bk-docker-xlpj/bazel-downstream-projects/upb/upb/bindings/lua/lua_proto_library.bzl", line 82, column 35, in _compile_upb_protos
		files = [_generate_output_file(ctx, name, "_pb.lua") for name in proto_sources]
	File "/var/lib/buildkite-agent/builds/bk-docker-xlpj/bazel-downstream-projects/upb/upb/bindings/lua/lua_proto_library.bzl", line 69, column 39, in _generate_output_file
		real_short_path = paths.relativize(real_short_path, package)
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/2056df990a8711c59c264304314251c8/external/bazel_skylib/lib/paths.bzl", line 182, column 13, in _relativize
		fail("Path '%s' is not beneath '%s'" % (path, start))
Error in fail: Path 'upb/bindings/lua/test.proto' is not beneath '/lua'
```

A manual bisect points to https://github.com/bazelbuild/bazel/commit/b33c5352fe59ded9dfcf10668bf104f977092218

@comius I know when you enabled the Starlark proto library, the downstream was green for upb, I guess there are new changes in upb that are broken by the Starlark proto library. Can you advise how to fix upb?

Wyverald:
> the downstream was green for upb, I guess there are new changes in upb that are broken by the Starlark proto library

[...]

PR #788, opened 2022-09-16
Use proto_common module in lua_proto_library

comius:
This handles all the paths correctly and reduces the code size. It should work correctly also internally.

Fixes: https://github.com/protocolbuffers/upb/issues/784

### DIFF
--- a/bazel/workspace_deps.bzl
+++ b/bazel/workspace_deps.bzl
@@ -21,6 +21,16 @@ def upb_deps():
     )
 
     maybe(
+        http_archive,
+        name = "rules_proto",
+        sha256 = "80d3a4ec17354cccc898bfe32118edd934f851b03029d63ef3fc7c8663a7415c",
+        strip_prefix = "rules_proto-5.3.0-21.5",
+        urls = [
+            "https://github.com/bazelbuild/rules_proto/archive/refs/tags/5.3.0-21.5.tar.gz",
+        ],
+    )
+
+    maybe(
         _github_archive,
         name = "com_google_protobuf",
         repo = "https://github.com/protocolbuffers/protobuf",
--- a/upb/bindings/lua/BUILD.bazel
+++ b/upb/bindings/lua/BUILD.bazel
@@ -35,6 +35,15 @@ load(
 
 licenses(["notice"])
 
+proto_lang_toolchain(
+    name = "lua_proto_toolchain",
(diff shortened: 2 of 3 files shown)


haberman:
Thanks for the change but we cannot do this yet. We need to support the two most recent Bazel LTS versions per our policy:

https://opensource.google/documentation/policies/cplusplus-support#3_build_systems

Until proto_common is supported in the latest two LTS releases, we cannot use it.

Is there a way to make this conditional on the availability of `proto_common`?

comius:
I'm reluctant to fix workarounds in upb, that are broken with fixes in proto_library for LTS release:  https://github.com/protocolbuffers/upb/issues/784

It doesn't sound like a good use of my time.

@meteocloudy, can we disable upb in downstream test and reenable it after LTS is released?

haberman:
I repeat my previous question:

> Is there a way to make this PR conditional on the availability of proto_common?

If we could use the new code-paths when a newer Bazel is available, we would not need to disable upb.

haberman:
For example, if we could test the Bazel version from Starlark, we could achieve both goals (support old Bazel versions and have less buggy behavior with newer Bazels).

comius:
Please hold this PR until after LTS.

haberman:
I don't think one LTS will be enough. We need to support the *two* most recent LTS releases. So we will need to wait until `proto_common` is in the last two LTS releases.

acozzette:
@comius We now require Bazel 6 or higher, so we can start relying on `proto_common`. However, we just moved upb into the protobuf [repo](https://github.com/protocolbuffers/protobuf/tree/main/upb) and I also can't merge this PR as it is due to the conflicts. If you can fix the conflicts and send us a PR for upb's new location then we would be happy to accept this, though.

meteorcloudy:
@acozzette Interesting to see upb is now a part of the protobuf repo. Do you think we should still keep the [upb module](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/upb) in the BCR for future versions?

acozzette:
@meteorcloudy Good question, I have not thought about this very much yet. For now, protobuf and upb are still in separate Bazel repos although they're in the same Git repo. Soon we will be merging the Bazel repos as well, though. So I guess we may want to keep the upb module in BCR for backward compatibility, but soon everyone should update their dependencies to just depend on protobuf instead.

PR #789, opened 2022-09-16
Make upb backwards and forwards compatible with Bazel 4.x, 3.5.x and LTS

comius:


meteorcloudy:
Hmm, upb still fails with Bazel@HEAD after this change: https://buildkite.com/bazel/bazel-auto-sheriff-face-with-cowboy-hat/builds/1039

comius:
The error message points to a different version of upb. Maybe we need to make upb green at head, so that last green gets updated?

meteorcloudy:
OK, thanks!

PR #836, opened 2022-10-06
Move generic comparison functions to upb/compare

cmburn:
Allows non-Python libraries to use the provided comparison functions. 
Moves them into `//upb/util:compare`


cmburn:
I opened a pull request about a month ago with this and accidentally closed it
immediately after. Deleted the branch entirely earlier when I was trying to
untangle some of my branches. Whoops!

PR #863, opened 2022-10-13
Update hash for 21.8 release

googleberg:
protobuf 21.8 release process: should point at latest commit on the prototbuf 21.x branch.


deannagarcia (approved):
Can you add the sha256 here as well? You should be able to see it when you run bazel build it will pop up at the top.

googleberg:
done

PR #980, opened 2022-11-28
Yet another attempt to apply the round-trip encoding fix.

ericsalo:
This time for sure!


haberman:
Once we rebase on top of https://github.com/protocolbuffers/upb/pull/1005, we should get all of the Bazel and Python tests too.

PR #1137, opened 2023-01-31
Build with incompatible_disallow_empty_glob

limdor:
In order to flip the flag, all downstream projects should be adapted. However, it is hard to fix them all if there are constant regressions. Adding it to the CI will ensure that once the project can build with incompatible_disallow_empty_glob it can keep building like that.
See: https://github.com/bazelbuild/bazel/pull/15327


acozzette:
Thank you for the pull request and sorry we never got around to reviewing this. We did end up recently adding `--incompatible_disallow_empty_glob` to the bazelrc, though.

PR #1169, opened 2023-02-10
Disable new c++20 ambiguous comparison warning

Alfus:


acozzette:
Thank you for the pull request and sorry for taking so long to get around to this. I don't think we want to accept this change since it completely turns off treating warnings as errors, but if you can find a more targeted way to fix the error then we would be more willing to accept that. We just moved upb into the main protobuf repo (github.com/protocolbuffers/protobuf) so feel free to send us a PR there.

PR #1237, opened 2023-03-22
Use requirements.txt in GHA tests

deannagarcia:
We recently saw a failure pass through where requirements.txt was updated in a way that was not compatible with some python versions. This updates the tests to directly use requirements.txt to download requirements to catch issues like this in the future.

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -43,6 +43,10 @@ jobs:
         with:
           name: python-wheels
           path: wheels/
+      - uses: actions/upload-artifact@v3
+        with:
+          name: requirements
+          paths: python/requirements.txt
 
   test_wheels:
     name: Test Wheels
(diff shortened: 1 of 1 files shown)


.github/workflows/python_tests.yml
```diff
@@ -43,6 +43,10 @@ jobs:
         with:
           name: python-wheels
           path: wheels/
+      - uses: actions/upload-artifact@v3
+        with:
+          name: requirements
+          path: python/requirements.txt
```
haberman, line 49: Where is the source of truth for this file? Where does the data come from? A comment would be helpful.

deannagarcia: What do you mean by that? The file is just being pulled out of the repository.

haberman: Oh I see. I thought maybe it was coming from a genrule().

Does it need to be uploaded at all then? Could we use a regular checkout action to obtain it?

haberman: Makes sense. A comment to that effect would be helpful. Otherwise I would assume that any artifact you are uploading was produced by the build process. It would help to know the rationale for why a source was being uploaded as an artifact.

deannagarcia: Makes sense, added a comment!

haberman (approved):

PR #1336, opened 2023-05-31
fix: add shebang in python interpreter

### LINKED ISSUE #1212: Unable to build on Alpine Linux
Trying to build the `//python/dist:binary_wheel` target on Alpine fails on this error:

```
Repository rule pip_repository defined at:
  /root/.cache/bazel/_bazel_root/9fabff63541097ed1dd428064ee5b4c7/external/rules_python/python/pip_install/pip_repository.bzl:474:33: in <toplevel>
ERROR: An error occurred during the fetch of repository 'pip_deps':
   Traceback (most recent call last):
        File "/root/.cache/bazel/_bazel_root/9fabff63541097ed1dd428064ee5b4c7/external/rules_python/python/pip_install/pip_repository.bzl", line 341, column 13, in _pip_repository_impl
                fail("rules_python failed: %s (%s)" % (result.stdout, result.stderr))
Error in fail: rules_python failed:  (src/main/tools/process-wrapper-legacy.cc:80: "execvp(/root/.cache/bazel/_bazel_root/9fabff63541097ed1dd428064ee5b4c7/external/system_python/interpreter, ...)": Exec format error
```

This is caused by the https://github.com/protocolbuffers/upb/commit/248ed86f2bfe6cf3dc7ffc69fa55cb488fd4a4ea commit, which creates an `interpreter` file containing `exec /.../python3 "$@"`. This file however is not executable as it is missing a shebang, hence the `exec format error` returned.

Adding a `#!/bin/sh` on top of this file fixes the issue, though I don't think this is the right fix.

vthib:
This fixes cases where the interpreter cannot be executed, as there is no indication about how to execute it.

Fixes #1212

### DIFF
--- a/bazel/system_python.bzl
+++ b/bazel/system_python.bzl
@@ -205,7 +205,7 @@ def _populate_package(ctx, path, python3, python_version):
         support = "Supported" if supported else "Unsupported",
     )
 
-    ctx.file("interpreter", "exec {} \"$@\"".format(python3))
+    ctx.file("interpreter", "#!/bin/sh\nexec {} \"$@\"".format(python3))
     ctx.file("BUILD.bazel", build_file)
     ctx.file("version.bzl", "SYSTEM_PYTHON_VERSION = '{}{}'".format(python_version[0], python_version[1]))
     ctx.file("register.bzl", _register.format(ctx.attr.name))


fowles (approved):

vthib:
I don't know what the failure is about, I don't have any conflicts on my side

fowles:
that is an internal thing for us.  Monday was  US holiday, but I am kicking it along now.

vthib:
@fowles is it possible to backport it to the 23.x branch? I realized it was not present in the recent release, and this is blocking me from building the project

fowles:
if you send a PR to cherry-pick this back to 23.x we can include it in the next point release.  Regardless of that, we are planning to do a 24.x release next week, so you may be happier just waiting.

vthib:
ah sounds good then i'll wait for the 24.x release. Thanks!

PR #1338, opened 2023-06-01
Fix `PyUpb_Message_MergeInternal` segfault

messense:
when `PyUpb_Message_MergeFromString` returns `NULL`, currently `PyUpb_Message_MergeInternal` will call `Py_DECREF` on `NULL` which results in a segmentation fault.

This patch switches to `Py_XDECREF` to fix the segfault.


fowles (approved):

ericsalo (approved):

acozzette:
@messense Thank you for the fix!

PR #1352, opened 2023-06-12
Export cmake config package files and add bootstrap buildings.

### LINKED ISSUE #1350: Export cmake CONFIG package.
It will be easier to use for cmake users when we export cmake CONFIG package.This package file will help users to resolve the dependencies and link order.

### LINKED ISSUE #1351: [Feature Request] Support bootstrap build upbc and lua binding with cmake.
I will be nice if we can build upbc and lua binding with cmake.

owent:
Fixes protocolbuffers/protobuf#13732 
Fixes protocolbuffers/protobuf#13731

+ Export cmake config package files.
+ Add bootstrap building support for cmake, and add stage0,stage1,stage2 targets just like in bazel.
+ Add `UPB_ENABLE_CODEGEN` options to decide whether to enable boostrap building and use upb,upbdefs and lupb in it to generate `*.upb.*`.

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1,19 +1,22 @@
 # This file was generated from BUILD using tools/make_cmakelists.py.
 
-cmake_minimum_required(VERSION 3.1)
-
-if(${CMAKE_VERSION} VERSION_LESS 3.12)
-    cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
-else()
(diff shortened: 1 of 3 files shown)


acozzette:
@owent Sorry for the bad news, but I don't think we will be able to accept this change. We just moved upb into the protobuf repo [here](https://github.com/protocolbuffers/protobuf/tree/main/upb) and we're also planning on doing a complete rewrite of upb's CMake config, so unfortunately we can't accept such a big change to the CMake config.

PR #1391, opened 2023-07-10
Expose methods needed by Ruby FFI using UPB_API

JasonLunn:
Once merged, the changes to ~`upb-ruby.h`~ `glue.c` in https://github.com/protocolbuffers/protobuf/pull/11483 can be reverted without breaking the FFI feature branch.

### DIFF
--- a/upb/collections/map.h
+++ b/upb/collections/map.h
@@ -120,16 +120,18 @@ UPB_API void upb_Map_SetEntryValue(upb_Map* map, size_t iter,
  */
 
 // Advances to the next entry. Returns false if no more entries are present.
-bool upb_MapIterator_Next(const upb_Map* map, size_t* iter);
+UPB_API_INLINE bool upb_MapIterator_Next(const upb_Map* map, size_t* iter);
 
 // Returns true if the iterator still points to a valid entry, or false if the
 // iterator is past the last element. It is an error to call this function with
 // kUpb_Map_Begin (you must call next() at least once first).
-bool upb_MapIterator_Done(const upb_Map* map, size_t iter);
+UPB_API_INLINE bool upb_MapIterator_Done(const upb_Map* map, size_t iter);
 
 // Returns the key and value for this entry of the map.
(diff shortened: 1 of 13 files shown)


upb/collections/map.h
```diff
@@ -120,16 +120,18 @@ UPB_API void upb_Map_SetEntryValue(upb_Map* map, size_t iter,
  */
 
 // Advances to the next entry. Returns false if no more entries are present.
-bool upb_MapIterator_Next(const upb_Map* map, size_t* iter);
+UPB_API_INLINE bool upb_MapIterator_Next(const upb_Map* map, size_t* iter);
```
haberman, line 123: `UPB_API_INLINE` is intended for inline functions. For non-inline functions, we should use `UPB_API`.

upb/hash/common.h
```diff
@@ -189,7 +189,7 @@ UPB_INLINE size_t upb_table_size(const upb_table* t) {
 
 UPB_INLINE bool upb_tabent_isempty(const upb_tabent* e) { return e->key == 0; }
 
-uint32_t _upb_Hash(const void* p, size_t n, uint64_t seed);
+UPB_API_INLINE uint32_t _upb_Hash(const void* p, size_t n, uint64_t seed);
```
haberman, line 192: Why is this needed? This is an implementation detail inside upb.

haberman (changes requested):

upb/hash/common.h
JasonLunn: `Message_Hash` (from `message.c`) and `Msgval_GetHash` (from `convert.c`) call it directly, so FFI needs to be able to do the same.

upb/collections/map.h
JasonLunn: Done

upb/hash/common.h
JasonLunn: Reverted

haberman (approved):

PR #1398, opened 2023-07-10
Updated CMake generated files.

haberman:
It seems that these are not currently updating manually.


mkruskal-google (approved):

acozzette:
Let me go ahead close this, because the CMakeLists.txt generation is fixed and we're winding down the use of this repo anyway.

PR #1449, opened 2023-08-04
Reuse pattern already in use for __has_attribute for __has_feature.

JasonLunn:


fowles (approved):

upb/port/undef.inc
```diff
@@ -63,6 +63,7 @@
 #undef UPB_UNPOISON_MEMORY_REGION
 #undef UPB_ASAN
 #undef UPB_ASAN_GUARD_SIZE
+#undef UPB_HAS_FEATURE
```
fowles, line 66: is UPB_HAS_ATTRIBUTE missing from this list?

JasonLunn: Yes, though it is undef'd [here](https://github.com/protocolbuffers/upb/blob/3d0119a2dd674851107f35f7c039dfcc1b0d2ee4/upb/port/def.inc#L219).

fowles: let's be consistent and undef these in the same place (your call which)

acozzette:
I'm going to close this as we discussed offline, since we have moved upb development into the protobuf repo and this PR is not quite ready to merge as-is.

PR #1458, opened 2023-08-07
Add weakref support to cpython upb message wrapper

bostikforever:
This adds weakref support to message objects in agreement with the 'python' and 'cpp' implementations.

See also: https://github.com/protocolbuffers/protobuf/issues/13727


acozzette:
We recently moved upb into the protobuf repo: https://github.com/protocolbuffers/protobuf/tree/main/upb I apologize for the inconvenience, but if you would still like to get this change merged then you will need to send us another pull request to that repo.

PR #1502, opened 2023-08-22
Fix lambda capture compiler error with c++20

kgreenek:
Connects to: https://github.com/protocolbuffers/protobuf/issues/13742

When compiling with C++20, the following error is produced:

```
upb/mini_table.hpp:63:22: note: add explicit 'this' or '*this' capture
upb/mini_table.hpp: In lambda function:
upb/mini_table.hpp:71:22: error: implicit capture of 'this' via '[=]' is deprecated in C++20 [-Werror=deprecated]
   71 |     return appender_([=](char* buf) {
```

In C++20, it is no longer allowed to implicitly capture 'this' in a lambda using [=].

This commit explicitly captures required values in the appropriate lambdas and removes all uses of [=] with lambdas.

### DIFF
--- a/upb/mini_descriptor/internal/encode.hpp
+++ b/upb/mini_descriptor/internal/encode.hpp
@@ -43,49 +43,51 @@ class MtDataEncoder {
   MtDataEncoder() : appender_(&encoder_) {}
 
   bool StartMessage(uint64_t msg_mod) {
-    return appender_([=](char* buf) {
+    return appender_([this, msg_mod](char* buf) {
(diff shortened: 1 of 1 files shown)


kgreenek:
Gentle bump on this

acozzette (approved):

acozzette:
@kgreenek Thank you! We recently moved upb into the main protobuf [repo](https://github.com/protocolbuffers/protobuf) but I will make sure to manually move this change there so that it doesn't get dropped.

PR #1532, opened 2024-03-29
Make it clearer that upb is archived

honglooker:
- moved all to archive/
- README now solely contains text about new development being done under protobuf/


haberman:
I think that moving the dotfiles to `archive/` also will avoid the CI failures.

honglooker:
ah nice - moved dotfiles

lmk whatcha think @haberman

haberman (approved):

